PR #18, opened 2015-03-10
When the app is started with wait-for-debugger request, it will wait unt...

PanayotCankov:
...il the debugger is connected and brake in the bootstrapper


fealebenpae:
I had made some comments on memory managing blocks, but I hadn't realized at the time that the TNSDebugging target is built with ARC enabled. Since those comments are inapplicable in this case I have removed them.

PR #38, opened 2015-03-19
Moving tns_modules inside app folder.

jasssonpet:


jasssonpet:
Ping @Fatme @atanasovg @dtopuzov

fealebenpae:
Since this is a breaking change for the CI builds and the CLI we'd like to receive confirmation of acknowledgement before we move forward with merging it.

This also means that the CI build for this patch will fail (as it has), even though this is a good patch.

jasssonpet:
Changed to be backwards compatible so it can be merged now.

PanayotCankov:
:+1:

PR #41, opened 2015-03-19
Enqueue microtasks with the common runloop modes

### LINKED ISSUE #39: Promise reactions have too low a priority on the runloop
Right now Promise .then()'s are preempted by scrolling and other UI events.

This is exhibited in the Cuteness app - images don't appear until scrolling ends because they are loaded with Promises.

/cc @hshristov

fealebenpae:
Promise reactions and other microtasks are executed on the current RunLoop in its default mode. Move this execution to the set of common modes to bump its priority.

Fixes #39


jasssonpet:
:+1:

PR #56, opened 2015-03-28
Remove debugging symbols in release.

jasssonpet:
This reduces the NativeScript.framework from 750MB to 50MB.


jasssonpet:
ping @dtopuzov @ErjanGavalji

PR #61, opened 2015-03-31
Add a global uncaught error handler

fealebenpae:
A general way to handle uncaught errors from the runtime. Expose the necessary APIs to let handlers spin up the inspector runloop.


jasssonpet:
:+1:

PR #63, opened 2015-04-01
Fix tagged pointer marshalling - close #23.

### LINKED ISSUE #23: Tagged pointers marshalling
Currently we crash when [tagged](http://en.wikipedia.org/wiki/Tagged_pointer) [pointers](https://www.mikeash.com/pyblog/friday-qa-2012-07-27-lets-build-tagged-pointers.html) are [marshalled to JavaScript](https://github.com/NativeScript/ios-runtime/blob/d143be3585f8bd328bf462dd1a3a138e4f01f5ed/src/NativeScript/NativeScript/ObjCTypes.mm#L47) as wrapper objects on iOS 7 - 64 bit. This affects roughly 3% of all devices ([statistics](http://david-smith.org/iosversionstats/)).

The reason lies in a bug in the Objective-C runtime, which seems to be fixed in iOS 8:

```objective-c
NSNumber *number = @123;
objc_setAssociatedObject(number, ...); // Crashes
```

Possible workarounds:
  - Use a `WeakGCMap` from pointer to `JSValue`

jasssonpet:


fealebenpae:
:+1:

PR #66, opened 2015-04-02
Add _getCurrentStack to help us save stacktrace log on crash

PanayotCankov:


fealebenpae:
Where is +[TNSRuntime _getCurrentStack] going to be used? If you already have access to an ExecState* you shouldn't use the _printCurrentStack code - it works best in the debugger only and is not thread-safe.

PanayotCankov:
We are experimenting with the native crashes in the ION.
For example:
```
var array = NSArray.alloc();
array.release();
array.release(); // Goes away with bad access
```
And we want to save the JS stack along the native stack using the kscrash report.

Ultimately we should be able to get the call stack of a thread with the native and JavaScript frames merged in a single report.

fealebenpae:
I'm afraid `+[TNSRuntime _printCurrentStack]` is not the correct method for this. It relies on `wtfThreadData()` which is thread-specific. Should this method be invoked on a thread other than the one that initialized the runtime, it'll produce a crash. I can't imagine what sort of a mess it'll create if that happens within a signal handler. Like I said in my first comment, `_printCurrentStack` carries implications that make it fit to be used solely while debugging.

The correct way to go about this with regards to the Companion app is to make its instance of `TNSRuntime` global. Then whenever any of its crash handlers fire on whichever thread, simply access the global runtime instance's VM and iterate its top call frame to get the JavaScript execution stack. I'm not sure if signal handlers pause all threads, but in case they don't you'll also need to lock the VM before walking its stack.

PanayotCankov:
We will prepare the release branch today and will iterate more on the call stacks at Monday.

PR #68, opened 2015-04-03
The hostname of the WebSocket server will be the same as the web host that servers the web inspector pages

PanayotCankov:
When a webserver on the device hosts the inspector,
the WebSocket client should connect to the device.
Currently it connects to "localhost" which is the Mac or PC from where it is accessed.


ivanbuhov:
:+1:

PR #71, opened 2015-04-06
Package.json's main property can be used to specify the bootstrap file

ivanbuhov:


jasssonpet:
:+1: 

I would prefer the logic for this to be in the template, and not in the require, but this is not a stopper for merging.

jasssonpet:
Related to https://github.com/NativeScript/NativeScript/issues/22

PR #80, opened 2015-04-14
Drop all locks when calling out to native

fealebenpae:
Turns out there's a nasty race condition waiting to happen if a thread other than the one that owns the VM tries to call into JavaScript using `FFICallback`. The guilty party is `ABAddressBookRequestAccessWithCompletion`, whose completion block is called on an arbitrary queue. This queue tries to acquire the VM's API lock, but fails because the VM owner thread is still holding on to it.

The answer is to temporarily drop all locks before calling a native API so that a worker thread spawned by the native API can have the chance to acquire the API lock.

### DIFF
--- a/src/NativeScript/NativeScript/FFICall.h
+++ b/src/NativeScript/NativeScript/FFICall.h
@@ -47,7 +47,8 @@ protected:
         return *static_cast<T*>(this->_arguments[index]);
     }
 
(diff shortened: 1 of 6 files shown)


jasssonpet:
:+1:

jasssonpet:
:+1:

PR #81, opened 2015-04-15
Metadata generator rewritten to C++

ivanbuhov:
Metadata generator rewritten to C++. 
Metadata is generated on build step of project template. 
There is no more metadata merger.


jasssonpet:
:+1:

PR #119, opened 2015-05-13
Change TNSDebugging's transport encoding

fealebenpae:
Graphite uses UTF-16 and thus its easier to change it here rather than break Graphite as well.


jasssonpet:
:+1:

jasssonpet:
:+1:

PR #131, opened 2015-05-15
Add interop.d.ts Fixed #93

### LINKED ISSUE #93: Interop TypeScript definitions does not allow `x instanceof interop.Pointer`
The `instaceof` operator should work and compile well with `interop.Pointer` and `interop.Reference`.
Possible solutions are to:
 - Avoid decomposing to interfaces and vars and simply use classes.
 - Export `var` for Pointer.

PanayotCankov:


jasssonpet:
We should also include `ReferenceType`, `FunctionReferenceType` and `BlockType`.

PR #138, opened 2015-05-19
Require modules with js extension - close #17.

### LINKED ISSUE #17: Requiring a file with extension not working
Requiring a file like

```javascript
var file = require("./folder/myfile.js");
```

is currently not working, while it is a prefectly valid scenario in Node:
[Node Modules](https://nodejs.org/api/modules.html)

This is working in the Android Runtime.

jasssonpet:


PanayotCankov:
:+1:

PR #140, opened 2015-05-20
Metadata API refactoring

ivanbuhov:
1. The old MetaFile and MetaFileReader structures are removed.
2. Intrduced new MetaFile and GlobalTable structures.
3. Map all parts of the binary metadata file to corresponding C structures, and access metadata information from methods and fields of the structures. This eliminates the need of global static MetaFileReader.
4. Top level module is exposed as string pointer field in Meta structure instead of a frameworkId (the introduction of frameworkId was made because of file size optimiziations but the benefit is insignificant and only the complexity was increased).


fealebenpae:
I think this patch has turned out very well. It makes the metadata code much easier to read and it very closely reflects the layout of the binary metadata.

Couple of things:
- Why the underscore in `TypeEncodingsList<T>::_count`? It's not a private member.
- Please refrain from C-style casts. Try to use `static_cast` and `reinterpret_cast` everywhere.

ivanbuhov:
@fealebenpae 
1. I submited a commit which removes all underscore prefixes from public struct fields. (bc66212ea9b42913fb6f3dba1b1cfdf9e528be87)
2. I submited a commit which transforms C-style casts to `reinterpret_cast` (2322a67fcb50f89052ec5c4648555002934034ca)
3. `TypeEncodingsList<T>::first()` actually doesn't expect the size of `count` to be 1. 
```
TypeEncoding* first() {
    return reinterpret_cast<TypeEncoding*>(this + 1);
 }
```
Here `this`'s type is `TypeEncoding *`, so `this + 1` performs pointer arithmetic, so the result pointer is pointing right after the TypeEncodingList structure in memory. Taking into account `sizeof(T)` or `sizeof(_count)` will enforce changes in  `TypeEncodingsList<T>::first()` method if a new field is added to the structure, so I think it is better to use the pointer arithmetic version.

fealebenpae:
@ivanbuhov D'oh, it appears I know nothing about pointer arithmetic :grin: Sorry about that.

You've still got a few `(char*)` casts in there, mostly when you need to work with offsets. I suggest replacing those with a utility function such as this:
```
template<typename To, typename From>
static To* offset_and_reinterpret(From* from, ptrdiff_t offset) {
	return reinterpret_cast<To*>(reinterpret_cast<intptr_t>(from) + offset);
}
```
Then `reinterpret_cast<T*>((char*)MetaFile::instance()->heap() + this->offset)` becomes `offset_and_reinterpret<T>(MetaFile::instance()->heap(), this->offset)`

Other than that, you've got :+1: from me, so feel free to merge.

PR #143, opened 2015-05-20
Insert richer modules information in metadata and expose it in metadata API.

ivanbuhov:
Each meta has a pointer to ModuleMeta structure which describes a top level module (only top level modules can be required). All ModuleMeta structures are also kept in global ModuleTable (an array of pointers to ModuleMetas, sorted by module name). This table is used to perform lookups by module name.


fealebenpae:
:+1:

Issue #96, opened 2015-04-28
iOS Extensions


PanayotCankov:
We need to build and distribute a shared framework out of the NativeScript.framework. This should be relatively easy, we have an internal proof of concept spike that proved that today views work with such framework. It should be distributed in the "tns-ios" npm package. However we should be sensitive for the filesize.

Then we need to provide bootstrappers and instructions how each type of the extension types can be set up manually. Please note that we will not integrate this in the CLI yet.

The most valuable deliverable would be a working WatchKit example. For integration with storyboards we will be ok to require manually written headers for now. In future if we provide our own XML markup for the WatchKit we will be able to generate headers from the XML. Or we can generate headers out of TypeScript declaration files if this would be more familiar to JavaScript developer.

We need to have only one copy of the metadata.

 - [x] [Metadata file should be initialized when the runtime is loading, not with static constructor](https://github.com/NativeScript/ios-runtime/issues/127)
 - [x] [The metadata file should contain extension availability information](https://github.com/NativeScript/ios-runtime/issues/128)
 - [x] Add shared NativeScript framework in the "tns-ios" package
 - [x] Add today view example
 - [x] [Add WatchKit example](https://github.com/NativeScript/ios-runtime/issues/120)
 - [x] Add article to describe how to enable NativeScript extensions by hand in Xcode

truongsinh:
I am sorry but I cannot quite get your idea. Is your "iOS extension" this https://developer.apple.com/app-extensions/ and this https://developer.apple.com/library/ios/documentation/General/Conceptual/ExtensibilityPG/ ?

jasssonpet:
@truongsinh Yes, the idea is to be able to write native iOS Extensions with NativeScript.

truongsinh:
+1 That is what we have been waiting for a loooong time!

jasssonpet:
You can see a simple working example here: https://github.com/NativeScript/ios-runtime/pull/110/files#diff-2

truongsinh:
Awesome, I though we have to wait, but apparently we can already play around it!

truongsinh:
I know I'm in the wrong repo, but I would like to ask is there a same feature for Android Widget in android-runtime repo? Issue reference would help!

jasssonpet:
ping @slavchev

slavchev:
@truongsinh I created an issue, see https://github.com/NativeScript/android-runtime/issues/69

roblav96:
[...]

PR #148, opened 2015-05-25
Add the NativeScript.framework to the tns-ios

PanayotCankov:
See #96

### DIFF
--- a/gruntfile.js
+++ b/gruntfile.js
@@ -134,6 +134,7 @@ module.exports = function (grunt) {
             packageComponents: {
                 files: [
                     { expand: true, cwd: "<%= outDistDir %>", src: ["NativeScript", "NativeScript/**"], dest: "<%= outPackageFrameworkDir %>" },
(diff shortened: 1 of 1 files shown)


ivanbuhov:
:+1:

fealebenpae:
Why is it copying the Debug build of the framework?

fealebenpae:
Could you please reopen this and change the build scripts to indicate that this is actually packaging the device+simulator build of the framework, rather than the debug build?

PanayotCankov:
We have Debug and Release folders where the Debug has the following architectures: i386, x86_64, armv7, arm64, and the Release has only the: armv7, arm64... since we can strip the simulator architectures on build, I do not think that we actually need 2 versions of the framework at all.

We will probably further polish the CMake build when @jasssonpet is back. You can log it if you are afraid we will forgot that.

For now I just need to complete the #96 iOS Extensions issue

jasssonpet:
We should remove the build with arm architectures only, because we now strip the non valid architectures on build.

PR #151, opened 2015-05-26
Release the proper value when new-ing classes

fealebenpae:
`ObjCConstructorCall` used to release the object returned by `alloc`, but it has to release the object returned by the initializer call instead, to account for cases when `alloc` returns a dummy placeholder factory object or when an initializer replaces `self`.


jasssonpet:
:+1: 

We could probably reuse `ConstructorResolutionTests.js`

PR #191, opened 2015-06-11
Cache require paths - close #139.

### LINKED ISSUE #139: Optimize require of JavaScript modules
We could try to make fewer file systems operations on each require if possible. This takes about 100ms of about 1000ms startup time on iPhone 5 at the moment.

An alternative is: https://github.com/NativeScript/nativescript-cli/issues/489

jasssonpet:
This reduces the Cuteness load time from 730ms to 630ms on iPhone 5s in Release.

### DIFF
--- a/src/NativeScript/require.js
+++ b/src/NativeScript/require.js
@@ -20,6 +20,7 @@
 
     var isDirectory = new interop.Reference(interop.types.bool, false);
 
+    var pathCache = new Map();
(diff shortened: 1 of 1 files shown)


PanayotCankov:
:+1: That makes somewhat 15% faster right?

enchev:
:+1:

enchev:
anyway .. I still like it! :)

PR #195, opened 2015-06-15
TNSDebugging should work with the dynamic NativeScript.framework.

PanayotCankov:
The JSStringCreateWithCFString symbol, used in TNSDebugging.h, seems to be stripped from the dynamic NativeScript.framework. Replaced it with JSStringCreateWithUTF8CString so that the debugging can work with that framework.


fealebenpae:
:+1:

PR #196, opened 2015-06-15
Handle memory pressure on iOS

### LINKED ISSUE #136: Handle memory pressure like WebProcess does
We should trigger GC when a low memory notification is recieved.

Remove this when done: https://github.com/NativeScript/ios-runtime/blob/36258e534faea66a6aa05a7f3f9e87cdeeef9b22/tests/TestRunner/app/ApiTests.js#L255

fealebenpae:
Close #136

### DIFF
--- a/src/NativeScript/TNSRuntime.mm
+++ b/src/NativeScript/TNSRuntime.mm
@@ -13,6 +13,10 @@
 #include <JavaScriptCore/JSGlobalObjectInspectorController.h>
 #include <JavaScriptCore/StrongInlines.h>
 
+#if PLATFORM(IOS)
+#import <UIKit/UIApplication.h>
+#endif
+
 #include "require.h"
 #include "inlineFunctions.h"
 #import "TNSRuntime.h"
@@ -22,7 +26,11 @@
 using namespace JSC;
 using namespace NativeScript;
 
-@implementation TNSRuntime
+@implementation TNSRuntime {
+#if PLATFORM(IOS)
+    id _memoryPressureNotificationSubscription;
+#endif
+}
 
 + (void)initialize {
     if (self == [TNSRuntime self]) {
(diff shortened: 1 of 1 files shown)


FireSt-dead:
:+1:

src/NativeScript/TNSRuntime.mm
```diff
@@ -36,6 +44,15 @@ - (instancetype)initWithApplicationPath:(NSString*)applicationPath {
         self->_applicationPath = [applicationPath copy];
         WTF::wtfThreadData().m_apiData = static_cast<void*>(self);
 
+#if PLATFORM(IOS)
+        VM* vm = self->_vm.get();
+        self->_memoryPressureNotificationSubscription = [[NSNotificationCenter defaultCenter] addObserverForName:UIApplicationDidReceiveMemoryWarningNotification object:nil queue:nil usingBlock:^(NSNotification *note) {
+            JSLockHolder lock(vm);
+            vm->heap.collect(FullCollection);
+            vm->heap.releaseDelayedReleasedObjects();
```
jasssonpet: Is this needed? It seems to me that `releaseDelayedReleasedObjects` will be called after the lock is released.

fealebenpae: I'm not sure this happens on every `unlock` - according to https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/runtime/JSLock.cpp#L162 `JSLock::unlock` will call `Heap::releaseDelayedReleasedObjects` only when the lock recursion is zeroed out, and I'm not certain this condition would be met every time the memory pressure notification fires.

Even if it turns out to be redundant, at worst a second call to  `Heap::releaseDelayedReleasedObjects` would be a no-op, so I suggest leaving it as it is.

Issue #199, opened 2015-06-19
Recursive method calls cause crash


FireSt-dead:
We have problems with the following code:
``` JavaScript
var RootViewController = UIViewController.extend({
    viewDidLoad: function() {
        UIViewController.prototype.viewDidLoad.apply(this, arguments);

        var label = new UILabel(CGRectMake(0, 0, 250, 60));
        label.text = "Hello, World!";

        label.center = this.view.center;
        label.textAlignment = NSTextAlignment.NSTextAlignmentCenter;

        UIView.animateWithDurationDelayOptionsAnimationsCompletion(3, 0, UIViewAnimationOptionCurveEaseIn, function() {
        	console.log("Set animation 1 properties");
            label.transform = CGAffineTransformMakeRotation(1);

        	UIView.animateWithDurationDelayOptionsAnimationsCompletion(3, 0, UIViewAnimationOptionCurveEaseOut, function() {
        		console.log("Set animation 2 properties");
        		UIView.setAnimationRepeatCount(3);
                label.alpha = 0.2;
        	}, function() {
        		console.log("Animation 2 Completed!");
        	})
        }, function() {
        	console.log("Animation 1 Completed!");
        });

        this.view.addSubview(label);
    }
});

var AppDelegate = UIResponder.extend({
    applicationDidFinishLaunchingWithOptions: function(application, launchOptions) {
        this._window = new UIWindow(UIScreen.mainScreen().bounds);
        this._window.backgroundColor = UIColor.whiteColor();
        this._window.rootViewController = new RootViewController();
        this._window.makeKeyAndVisible();
        return true;
    }
}, {
    protocols: [ UIApplicationDelegate ]
});

UIApplicationMain(0, null, null, NSStringFromClass(AppDelegate.class()));
```

PR #200, opened 2015-06-22
Recursive method calls

PanayotCankov:
Fix for #199

### DIFF
--- a/src/NativeScript/Calling/FFICall.cpp
+++ b/src/NativeScript/Calling/FFICall.cpp
@@ -8,8 +8,93 @@
 
 #include "FFICall.h"
 
+const size_t MAX_KNOWN_ALIGNMENT = 8;
+
 namespace NativeScript {
 using namespace JSC;
+
+FFICall::FFICallFrame::FFICallFrame(FFISignature& signature, ExecState* execState, Byte* buffer)
+    : _execState(execState) {
+
+    _result = buffer + signature.resultOffset;
+    _arguments = reinterpret_cast<void**>(buffer + signature.argsArrayOffset);
+
+    for (size_t i = 0; i < signature.argsCount; ++i) {
+        _arguments[i] = buffer + signature.argValueOffsets[i];
+    }
+}
+
+FFICall::FFISignature::FFISignature()
+    : argValueOffsets(nullptr)
+    , postCalls(nullptr) {
+}
+
+void align_offset(size_t& offset, size_t alignment) {
+    size_t misaligned = offset % alignment;
+    if (misaligned) {
+        offset += alignment - misaligned;
+    }
+}
+
+size_t allocate_aligned_offset(size_t& offset, size_t size, size_t alignment) {
+    if (!size) {
+        return offset;
+    }
+    align_offset(offset, alignment);
+    size_t result = offset;
+    offset += size;
+    return result;
+}
+
+size_t allocate_aligned_ffi_arg(size_t& offset, ffi_type* type) {
+    return allocate_aligned_offset(offset, type->size, type->alignment);
+}
+
+void FFICall::FFISignature::initialize(ffi_cif* cif, WTF::Vector<FFITypeMethodTable> parameterTypes, size_t initialArgumentIndex) {
+    argsCount = cif->nargs;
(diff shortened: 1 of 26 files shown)


ivanbuhov:
:+1:

jasssonpet:
Are there any performance implications in the startup time after this fix?

fealebenpae:
I think this is too complex and roundabout solution to the problem.

The crux of the matter is that the cleanup logic in `postCall` of the various FFI type objects breaks with recursion because we reuse the FFI argument buffers between calls to minimize allocations. I propose to deal away with `postCall` completely and use nest-able release pools, similar to Objective-C's `NSAutoreleasePool`, to handle cleanup. It appears you have chosen to do away with the buffer reuse instead.

I maintain that release pools would be a more elegant change with far more deletions than insertions that would make the FFI paths simpler, but let's examine this patch on its own merits.

We accept, for performance's sake, that we need to minimize the number of allocations during each FFI call, and indeed this has led to the current state of argument buffer reuse. In light of that, it's a good idea to not allocate N buffers of varying size, where N is the number of arguments needed for the FFI call, but trying to align those allocations is pointless. The `ffi_call` API constructs its own call frame internally, with proper alignments and everything. However, this only happens on x86, because of the calling conventions endemic to that architecture. Our performance target however is ARM, and the arm-apple-darwin conventions place arguments on CPU registers where possible. An alternative would be to use the `ffi_raw_call` API, which accepts not a vector of argument buffers but a properly populated and aligned call frame. However, on ARM SYSV this falls back to `ffi_call` internally, because that architecture/platform combo does not pass arguments on the stack in the same way x86 does. This means that all the custom alignment logic can be safely omitted and the arguments can be placed in a packed buffer.

I do not like the hand-rolled linked-list approach you have chosen in `FFISignature` to hold the `PostCall` structures. I am not certain why a linked list here is better than a plain vector or even a simple null-terminated array, but if you absolutely need to use one, it's better to go with `std::list` or `std::forward_list` instead of rolling your own.

You've also removed some of the `execState->hadException()` checks which caused FFI calls to fail safely before calling out to native in case a type has thrown an error while marshaling its argument value.

Again, I am strongly in favor of the release pool approach, especially as it's implemented in the winrt prototype. Is there a technical reason you've chosen to eliminate buffer reuse instead of going with release pools?

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -8,9 +8,94 @@
...
+        _arguments[i] = buffer + signature.argValueOffsets[i];
+    }
+}
+
+FFICall::FFISignature::FFISignature()
+    : argValueOffsets(nullptr)
+    , postCalls(nullptr) {
+}
+
+void align_offset(size_t& offset, size_t alignment) {
```
jasssonpet: You can use `static` (or anonymous C++ namespaces) to make functions and global variables internal to the file.

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -8,9 +8,94 @@
 
 #include "FFICall.h"
 
+const size_t MAX_KNOWN_ALIGNMENT = 8;
+
 namespace NativeScript {
 using namespace JSC;
 
+FFICall::FFICallFrame::FFICallFrame(FFISignature& signature, ExecState* execState, Byte* buffer)
```
jasssonpet: Instead of `Byte` you can use `int8_t` or `uint8_t` like in the other files.

src/NativeScript/Calling/FFICall.h
```diff
@@ -33,36 +61,72 @@ class FFICall : public JSC::InternalFunction {
 
     static void visitChildren(JSC::JSCell*, JSC::SlotVisitor&);
 
-    void preCall(JSC::ExecState*);
+    template <class Derived>
+    static JSC::EncodedJSValue baseExecuteCall(JSC::ExecState* execState) {
+        auto instance = JSC::jsCast<Derived*>(execState->callee());
 
-    JSC::JSValue postCall(JSC::ExecState*);
+        Byte* buffer = (Byte*)alloca(instance->_signature.stackSize);
```
jasssonpet: You can use `static_cast` here.

PR #201, opened 2015-06-22
Add application that uses the updated webkit webview to show web inspector frontend

KristinaKoeva:


KristinaKoeva:
run ci

fealebenpae:
Let's flesh out the CLI integration and make sure everything works on both ends and then merge this change.

PR #207, opened 2015-07-02
Removed postCall-s from type method tables. Add ReleasePool to handle…

PanayotCankov:
… block deallocations.

### DIFF
--- a/src/NativeScript/Calling/FFICall.cpp
+++ b/src/NativeScript/Calling/FFICall.cpp
@@ -92,8 +92,4 @@ void FFICall::preCall(ExecState* execState, uint8_t* buffer) {
         }
     }
 }
-
-JSValue FFICall::postCall(ExecState* execState, uint8_t* buffer) {
-    return this->_returnType.read(execState, buffer + this->_returnOffset, this->_returnTypeCell.get());
-}
 }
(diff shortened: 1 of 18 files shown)


src/NativeScript/Runtime/ReleasePool.h
```diff
@@ -0,0 +1,88 @@
+#ifndef __NativeScript__ReleasePool__
```
jasssonpet: We could add the license section to be consistent with other headers.

fealebenpae:
This look good. You can also update the UTF8CString type's [`write`](https://github.com/NativeScript/ios-runtime/blob/master/src/NativeScript/Marshalling/Fundamentals/FFIPrimitiveTypes.cpp#L135) function to take advantage of release pools.

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -94,17 +94,6 @@ void FFICall::preCall(ExecState* execState, uint8_t* buffer) {
...
-        void* argumentBuffer = reinterpret_cast<void**>(buffer + this->_argsArrayOffset)[i + this->_initialArgumentIndex];
-        JSCell* parameterType = this->_parameterTypesCells[i].get();
-        this->_parameterTypes[i].postCall(execState, argument, argumentBuffer, parameterType);
-
-        if (execState->hadException()) {
-            return jsUndefined();
-        }
-    }
-
     return this->_returnType.read(execState, buffer + this->_returnOffset, this->_returnTypeCell.get());
```
fealebenpae: Does the compiler properly inline calls to `FFICall::postCall`? Since it's such a small method now we may want it to be inlined. Perhaps we can move the body of `postCall` to the header so it can be inlined?

fealebenpae:
:+1:

PR #208, opened 2015-07-02
Throw NSError as JS error

### LINKED ISSUE #186: Transform (NSError **) parameters to JavaScript error throw
Obj-C methods which return `BOOL` and have(`NSError **`/`CFErrorRef*`) as last parameter should be exposed as methods without the last parameter and have to throw JavaScript error on failure (the same convention is used in Swift 2).

PRs:
* https://github.com/NativeScript/ios-runtime/pull/208
* https://github.com/NativeScript/ios-metadata-generator/pull/8
* https://github.com/NativeScript/ios-metadata-generator/pull/9
* https://github.com/NativeScript/ios-runtime-docs/pull/33
* https://github.com/NativeScript/NativeScript/pull/404

Subtasks:
* [x] Research Swift 2 projection
* [x] Discuss possible NativeScript projections
* [x] Implement calling of such methods
* [x] Implement overriding of such methods
* [x] Implement exposing of such methods
* [x] Implement swizzling of such methods
* ~~Implement constructor calls of such initializers~~
* [x] Update TypeScript definitions
* [x] Write tests for all cases
* [x] Update documentation
* [x] Make PR in modules to use the new features

jasssonpet:
Close #186

### DIFF
--- a/src/NativeScript/ObjC/ObjCMethodCall.h
+++ b/src/NativeScript/ObjC/ObjCMethodCall.h
@@ -65,6 +65,8 @@ private:
 
     bool _retainsReturnedCocoaObjects;
 
+    bool _hasErrorOutParameter;
+
     SEL _selector;
 };
 }
--- a/src/NativeScript/ObjC/ObjCMethodCall.mm
+++ b/src/NativeScript/ObjC/ObjCMethodCall.mm
@@ -28,6 +28,14 @@ void ObjCMethodCall::finishCreation(VM& vm, GlobalObject* globalObject, const Me
 
     Base::initializeFFI(vm, returnTypeCell, parameterTypesCells, 2);
     this->_retainsReturnedCocoaObjects = metadata->ownsReturnedCocoaObject();
+    this->_hasErrorOutParameter = metadata->hasErrorOutParameter();
+
+    if (this->_hasErrorOutParameter) {
+        this->_argumentCountValidator = [](FFICall* call, ExecState* execState) {
+            return execState->argumentCount() == call->parametersCount() ||
+                   execState->argumentCount() == call->parametersCount() - 1;
+        };
+    }
 
     this->_msgSend = reinterpret_cast<void*>(&objc_msgSend);
     this->_msgSendSuper = reinterpret_cast<void*>(&objc_msgSendSuper);
@@ -63,6 +71,13 @@ EncodedJSValue ObjCMethodCall::derivedExecuteCall(ExecState* execState, uint8_t*
     id target = NativeScript::toObject(execState, execState->thisValue());
     Class targetClass = object_getClass(target);
 
+    NSError* outError = nil;
+    if (this->_hasErrorOutParameter) {
+        if (this->_parameterTypesCells.size() - 1 == execState->argumentCount()) {
+            this->setArgument(buffer, this->_argsCount - 1, &outError);
+        }
+    }
+
     if (class_conformsToProtocol(targetClass, @protocol(TNSDerivedClass))) {
         objc_super super = { target, class_getSuperclass(targetClass) };
 #ifdef DEBUG_OBJC_INVOCATION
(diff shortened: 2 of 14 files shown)


src/NativeScript/ObjC/ObjCMethodCall.mm
```diff
@@ -19,6 +22,22 @@
 
 const ClassInfo ObjCMethodCall::s_info = { "ObjCMethodCall", &Base::s_info, 0, CREATE_METHOD_TABLE(ObjCMethodCall) };
 
+static bool hasErrorOutParameter(ExecState* execState, const WTF::Vector<JSCell*>& parameterTypesCells) {
+    if (!(parameterTypesCells.size() > 0)) {
```
fealebenpae: I believe you can use `WTF::Vector<T>::isEmpty()` here.

src/NativeScript/ObjC/ObjCMethodCall.h
```diff
@@ -46,7 +46,8 @@ class ObjCMethodCall : public FFICall {
 
 private:
     ObjCMethodCall(JSC::VM& vm, JSC::Structure* structure)
-        : Base(vm, structure) {
+        : Base(vm, structure)
+        , _hasErrorOutParameter{ false } {
```
fealebenpae: I could be wrong, but I believe we don't have to explicitly initialize this to `false`.

fealebenpae:
Looks good to me. Simple and elegant. Will you be handling `ObjCMethodCallback` next?

src/NativeScript/ObjC/ObjCMethodCall.h
jasssonpet: [C++11: 5.3.4/15]: A new-expression that creates an object of type T initializes that object as follows: If the new-initializer is omitted, the object is default-initialized (8.5); if no initialization is performed, the object has indeterminate value.

jasssonpet: You are right, we always set it anyway in `finishCreation` call.

jasssonpet:
@fealebenpae yes

src/NativeScript/ObjC/ObjCMethodCall.h
fealebenpae: I don't really mind initializing it in the constructor, I just feel queasy from the curly bracket syntax. I'm used to the round brackets syntax we employ elsewhere in the codebase :grinning:

jasssonpet:
Moved the `NSError**` last parameter check to https://github.com/NativeScript/ios-metadata-generator/tree/jasssonpet/method-nserror-out-param-flag

jasssonpet:
@fealebenpae Could you take a look at the `MethodCallback` changes in the second commit.

src/NativeScript/ObjC/ObjCMethodCallback.mm
```diff
@@ -46,11 +71,23 @@
...
+
+    if (methodCallback->_hasErrorOutParameter && execState->hadException()) {
+        size_t methodCallbackLength = jsDynamicCast<JSObject*>(methodCallback->function())->get(execState, execState->vm().propertyNames->length).toUInt32(execState);
+        if (methodCallbackLength == methodCallback->parametersCount() - 1) {
+            JSValue exception = execState->exception();
+            execState->clearException();
+            memset(retValue, 0, methodCallback->_returnType.ffiType->size);
+            id marshalledException = NativeScript::toObject(execState, exception);
+            NSError* nserror = [NSError errorWithDomain:@"TNSErrorDomain" code:164 userInfo:marshalledException];
+            **(static_cast<id**>(argValues[methodCallback->parametersCount() + methodCallback->_initialArgumentIndex - 1])) = nserror;
```
jasssonpet: I'll add a check if the pointer is not `nil`.

fealebenpae:
I wonder if we can do something smarter in cases where a method returns `BOOL` and accepts `NSError**` - what do you think about making their return type `void` in JavaScript and setting their return value in `ObjCMethodCallback` based on whether or not the user code threw an error? It seems to me that in this pattern the return value is inextricably linked to the out parameter. If we hide the out parameter and project it with JavaScript throw mechanics the boolean return value makes little sense on its own, both in calling native methods and overriding them.

jasssonpet:
@fealebenpae Made the `ObjCMethodCallback` with `BOOL` return type to return `YES` if no error is thrown.

PR #209, opened 2015-07-02
Aligned FFICall arguments.

PanayotCankov:


fealebenpae:
Excellent. I had not considered our reliance on buffer alignment when I commented on #200, so it's better to be safe than sorry and align.

We should support all cases of reentrant recursion with none-to-negligible slow downs on calling, right? Actually, with the refactoring and inlining and general code stripping in this series of changes I believe we ought to be even the tiniest bit faster on each call to native.

Good work, the new execution flow enables a number of possible future improvements to the FFI layer.

PanayotCankov:
Thanks :)

With the buffer allocated on the stack it shouldn't be much slower. That type of allocation should be fast as hell. We do some additional calculations in "initialize" but they are per method and work with O(n) anyway. As for an actual call, the only thing I could think of that may be slower is access to argument pointers that now uses "bufferPointer + argmentOffset" to calculate the arugmentPointer.

As a room for improvement, **if** we unique the signatures of the methods in the metadata, we can reincarnate the [FFICall::FFISignature](https://github.com/NativeScript/ios-runtime/blob/recursive-method-calls/src/NativeScript/Calling/FFICall.h#L103) and calculate the buffer size and offsets once per unique signature instead of once per method.

Also it would be worth checking the callbacks. They may have similar issues. For example if a method returns a block and call itself recursively.

fealebenpae:
I believe @ivanbuhov mentioned some time ago plans to unique objects in our API metadata container, which would unique instances of `Metadata::TypeEncodingsList`. We could then cache a lot of the FFI machinery between methods with the same signature, including calculated offsets and `ffi_cif`s, using the uniqued type encodings as keys.

ivanbuhov:
@fealebenpae I made an issue for this - https://github.com/NativeScript/ios-runtime/issues/216. This is a good candidate for my next task.

PR #210, opened 2015-07-06
Add test for recursively calling methods that return block

PanayotCankov:


fealebenpae:
This test doesn't exercise the bridge. You need to call the overridden method in Objective-C, not in JavaScript.

PanayotCankov:
@fealebenpae I am sorry, we need to add tags for the pull requests whether they are there to just trigger builds, or are actually ready for a review :(

fealebenpae:
It's not that it's wrong, but making `JSBlock` a class and overloading the `new` operator looks weird to me. You could keep it a struct, `malloc` it, and then in `disposeBlock` you can just call `JSC::Strong<T>::clear()` instead of going through creating a whole class with a destructor just to clear out the handle.

PanayotCankov:
Well, I like classes.

Technically they do not differ much from structures. But wrapping this in a separate class allows me to pack JSBlock's internals together. When you take a look at it, it requires just a few additional lines of code, but in return, it is now pretty obvious that the JSBlockDescriptor and the BLOCK_NEEDS_FREE won't be used anywhere else but in the JSBlock. It feels to me this class is like the thing that keeps the [Pudge](https://encrypted-tbn2.gstatic.com/images?q=tbn:ANd9GcR_DyrA7RgRVTJqoOi09mUi-st7-myInmSd0AhJ4C0fRgsYbzEK)'s composite body from falling apart.

Anyway, if you still don't like it I will revert it and simply "JSC::Strong<T>::clear()" in the disposeBlock. What do you say?

fealebenpae:
I understand, and packing things in classes is certainly a valid form of code organization. My concern is that the code moves away from block literals as found in libclosure.

How about the following: we can keep revert it to a structure, but keep the static members such as `copy`, `dispose` and `create` as well as enum declarations within it, while getting rid of access modifiers, the `new` overload and the constructor/destructor pair. That way `JSBlock` will still look much more like a block literal, while still encapsulating its logic and helpers.

PanayotCankov:
Ok, that sounds reasonable. I will branch with a new commit in a while.

PR #212, opened 2015-07-06
Link BlankApp project with TestFixtures.

jasssonpet:
@fealebenpae Any better way to do this?


fealebenpae:
How about we create a new macro, say `LinkWithTestFixtures`, that accepts a target, possibly created by `CreateNativeScriptApp`, and adds the necessary dependencies and link flags?

jasssonpet:
Shouldn't the include directories be set before the target is created.

fealebenpae:
[`target_include_directories`](http://www.cmake.org/cmake/help/v3.0/command/target_include_directories.html)?

jasssonpet:
That's just what I was looking for, thanks :)

jasssonpet:
@fealebenpae This PR is working on my machine. We should fix the PR build to reconfigure WebKit as needed.

jasssonpet:
run ci

cmake/LinkTestFixtures.cmake
```diff
@@ -0,0 +1,14 @@
+function(LinkTestFixtures _target)
+    set(TESTFIXTURES_DIR "${CMAKE_SOURCE_DIR}/tests/TestFixtures")
+
+    file(READ "${TESTFIXTURES_DIR}/exported-symbols.txt" TESTFIXTURES_EXPORTED_SYMBOLS)
```
fealebenpae: I did a little poking around, and it appears `file(STRINGS "${TESTFIXTURES_DIR}/exported-symbols.txt" TESTFIXTURES_EXPORTED_SYMBOLS)` can read the contents of a the exported symbols list file and split its lines into a list. Can we use it instead?

jasssonpet: Done

PR #222, opened 2015-07-14
Implement and consume NSFastEnumeration

fealebenpae:
This change exposes the Objective-C fast enumeration machinery by adding a `Symbol.iterator` property to the prototype of classes that implement fast enumeration.

This allows code expressed with a `for...in` statement in Objective-C to be implemented with JavaScript's `for...of` statement in NativeScript. Also, JavaScript APIs that expect iterable objects can now consume `NSFastEnumeration`-conforming Objective-C objects.

What's more, `TNSArrayAdapter` now implements bulk enumeration to speed up fast enumeration when the adapted object is the array exotic.

### DIFF
--- a/src/NativeScript/ObjC/TNSArrayAdapter.mm
+++ b/src/NativeScript/ObjC/TNSArrayAdapter.mm
@@ -48,4 +48,29 @@ using namespace JSC;
     return toObject(self->_execState, self->_object.get()->get(self->_execState, index));
 }
 
+- (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState *)state objects:(id [])buffer count:(NSUInteger)len {
+    if (state->state == 0) { // uninitialized
+        state->state = 1;
+        state->mutationsPtr = reinterpret_cast<unsigned long*>(self);
+        state->extra[0] = 0; // current index
+        state->extra[1] = self->_object->get(self->_execState, self->_execState->propertyNames().length).toUInt32(self->_execState);
+    }
+
+    NSUInteger currentIndex = state->extra[0];
+    unsigned int length = state->extra[1];
+    NSUInteger count = 0;
+    state->itemsPtr = buffer;
+
+    JSLockHolder lock(self->_execState);
+    while (count < len && currentIndex < length) {
+        *buffer++ = toObject(self->_execState, self->_object->get(self->_execState, currentIndex));
+        currentIndex++;
+        count++;
+    }
+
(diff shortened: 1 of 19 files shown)


tests/TestRunner/app/ApiTests.js
```diff
@@ -432,4 +432,17 @@ describe(module.id, function () {
             expect(NSDate.dateWithTimeIntervalSinceReferenceDate(0).class()).toBe(NSDate);
         });
     }
+    
+    it("should be able to iterate over NSFastEnumeration objects", function () {
```
jasssonpet: I'll suggest trying with a bigger array (e.g. 30/32 elements), so the case where the JSIterator buffer is filled and moved to be covered.

fealebenpae: `NSArray` always returns its inner pointerand never uses the buffer provided. What I can do is add another test that uses a big array and iterates over its 'reverseObjectsEnumerator' which does copy items to the buffer provided.

jasssonpet: That'll be great :)

jasssonpet:
:+1:

src/NativeScript/ObjC/TNSArrayAdapter.mm
```diff
@@ -48,4 +48,38 @@ - (id)objectAtIndex:(NSUInteger)index {
...
+            state->state = 1;
+            state->mutationsPtr = reinterpret_cast<unsigned long*>(self);
+            state->extra[0] = 0; // current index
+        } else {
+            state->state = 2;
+        }
+    }
+
+    if (state->state == 2) { // we don't want to bother with bulk enumeration for regular objects right now, so fall back to super
+        return [super countByEnumeratingWithState:state objects:buffer count:len];
```
jasssonpet: I'll suggest to move this check at the top, because modifying the state will break the super logic.

fealebenpae:
ping @jasssonpet

jasssonpet:
:+1:

fealebenpae:
@jasssonpet: I felt like adding a `Symbol.iterator` projection when extending an Objective-C class in JavaScript. With this I feel that the `NSFastEnumeration` projection is complete.

jasssonpet:
Great :+1:, we can update the documentation, so that our users can use this new feature :)

fealebenpae:
And I should update the metadata generator as well in order to add TypeScript declarations for the new `Symbol.iterator` methods.

Issue #189, opened 2015-06-11
Expose notifications when a NS app goes to background/foreground


KristinaKoeva:
Those notifications would help CLI/AppBuilder determine when a debugging session could be made

PR #223, opened 2015-07-15
Add notifications so that clients will know when the app is going to resign or become active

KristinaKoeva:
https://github.com/NativeScript/ios-runtime/issues/189

### DIFF
--- a/src/debugging/TNSDebugging.h
+++ b/src/debugging/TNSDebugging.h
@@ -5,6 +5,7 @@
 //  Copyright (c) 2015 г. Telerik. All rights reserved.
 //
 
+#import <UIKit/UIApplication.h>
 #import <JavaScriptCore/JavaScript.h>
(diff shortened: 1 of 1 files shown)


fealebenpae:
Please wrap the UIKit-specific code in `#if PLATFORM(IOS)` blocks. Hopefully this will run on OS X before the turn of the next millennium.

KristinaKoeva:
We are probably going to need a different project template for OS X. Moreover ```#if PLATFORM(IOS)``` is defined in ```wtf\\Platform.h``` which is not available for the template. We could use ```#if TARGET_OS_IPHONE``` but as you saw the logic would become a little bit complicated so if you don't mind I will leave it as it is.

KristinaKoeva:
run ci

fealebenpae:
@KristinaKoeva: You're right, we could end up needing a separate template for OS X apps anyway.

PR #224, opened 2015-07-15
Buhov/memory management of class clusters

ivanbuhov:


src/NativeScript/GlobalObject.mm
```diff
@@ -41,11 +42,31 @@
...
+    Class klass = constructor->klass();
+    id instance = [klass alloc];
+    GlobalObject* globalObject = jsCast<GlobalObject*>(execState->lexicalGlobalObject());
+
+    if (ObjCConstructorDerived* constructorDerived = jsDynamicCast<ObjCConstructorDerived*>(constructor)) {
+        [instance release];
+        JSValue jsValue = toValue(execState, instance, ^{ return constructorDerived->instancesStructure(); });
+        return JSValue::encode(jsValue);
+    } else if (jsDynamicCast<ObjCConstructorNative*>(constructor) != nullptr) {
+        AllocatedPlaceholder* allocatedPlaceholder = AllocatedPlaceholder::create(execState->vm(), jsCast<GlobalObject*>(execState->lexicalGlobalObject()), AllocatedPlaceholder::createStructure(execState->vm(), globalObject, constructor->instancesStructure()->storedPrototype()), instance, constructor->instancesStructure());
```
fealebenpae: Rather than creating a new instance of `AllocatedPlaceholder`'s `JSC::Structure` each time, let's keep one instance of the structure per class in `ObjCConstructorNative` and reuse that, just like we do for `ObjCWrapperObject`'s structure.

tests/TestRunner/app/Marshalling/ObjCTypesTests.js
```diff
@@ -178,4 +178,27 @@ describe(module.id, function () {
     it("NSNull", function () {
         expect(NSNull.null() instanceof NSObject).toBe(true);
     });
+         
+    it("AllocatedPlaceholderHasOnlyOneSymbol", function () {
```
fealebenpae: Why do we need this test? It only begs to be broken if we add another symbol in the future.

src/NativeScript/AllocatedPlaceholder.h
```diff
@@ -0,0 +1,48 @@
...
+    id _wrappedObject;
+
+    AllocatedPlaceholder(JSC::VM& vm, JSC::Structure* structure)
+        : Base(vm, structure) {
+    }
+
+    void finishCreation(JSC::VM& vm, GlobalObject* globalObject, id wrappedObject, JSC::Structure* instanceStructure) {
+        Base::finishCreation(vm);
+        this->_wrappedObject = wrappedObject;
+        this->putDirect(vm, globalObject->instanceStructureIdentifier(), JSC::JSValue(instanceStructure), JSC::ReadOnly | JSC::DontEnum | JSC::DontDelete);
```
fealebenpae: You don't have to wrap `JSCell*` values in a `JSValue` call, because `JSValue` already has an implicit constructor that accepts `JSCell`. It's purely a cosmetic nitpick, but the rest of the codebase does not explicitly construct `JSValue` this way, so I suggest removing the explicit calls to the `JSValue` constructor in this patch.

PR #234, opened 2015-07-21
JavaScript functions, when passed to Objective-C as blocks, will now round-trip back to JavaScript, as the initial function object.

PanayotCankov:


src/NativeScript/ObjC/Block/ObjCBlockType.mm
```diff
@@ -88,7 +96,14 @@ static void disposeBlock(JSBlock* block) {
 JSValue ObjCBlockType::read(ExecState* execState, const void* buffer, JSCell* self) {
     GlobalObject* globalObject = jsCast<GlobalObject*>(execState->lexicalGlobalObject());
     ObjCBlockType* blockType = jsCast<ObjCBlockType*>(self);
-    return ObjCBlockCall::create(execState->vm(), globalObject->objCBlockCallStructure(), *static_cast<const id*>(buffer), blockType);
+    id block = *static_cast<const id*>(buffer);
+
+    JSValue objCBlockCallback = JSBlock::getJSFunction(block);
```
jasssonpet: You can put the assignment in the if statement to reduce its scope.

tests/TestFixtures/TNSTestNativeCallbacks.h
```diff
@@ -12,6 +12,10 @@
 
 @interface TNSTestNativeCallbacks : NSObject
 
+typedef void (^TNSBlock)();
+
+@property(assign, nonatomic) TNSBlock block;
```
jasssonpet: I would suggest to move this to `TNSObjCTypes.m` where we have tests like:
* `methodWithNSDate`
* `methodWithNSArray`
* `methodWithNSDictionary`
* `methodWithNSData`

which test the same thing (that the object is the same).

Also removing the property from the header won't make the test fail, an is a bit brittle.

fealebenpae: > imo JS is brittle.

FTFY :grinning: 

We can add an expectation that the `block` property exists with the `in` operator, but what you've touched upon is a problem everywhere, not just in this test.

jasssonpet: That's why we try to avoid accessing native properties in our tests.

fealebenpae:
run ci

fealebenpae:
The build currently fails with:
`src/NativeScript/ObjC/Block/ObjCBlockType.mm:78:28: error: cannot cast from type 'id' to pointer type 'NativeScript::JSBlock *'`
`reinterpret_cast` should fix it.

Issue #231, opened 2015-07-20
Expose a conversion between ArrayBuffers and NSData


PanayotCankov:
In some scenarios where Objective-C APIs are used to implement JavaScript specs we have provide utilities to easily, with low performance overhead and low memory footprint:

 1. Create ArrayBuffers from NSData
 2. Create NSData from ArrayBuffers

Currently our marshaling supports the second scenario, when ArrayBuffer object is passed as argument somewhere NSData is expected it gets wrapped by the marshaling service. We should check if this is sufficient.

NathanaelA:
Do you actually have code that does the conversion between nsdata to ArrayBuffer?   I'm actually running into this where I'm getting a nsdata object and can't do anything with it.  If you point me to where you are creating the arraybuffer i might be able to reuse that code to get the data into js in a usable form.

fealebenpae:
You can check my pull request https://github.com/NativeScript/ios-runtime/pull/235

fealebenpae:
Implemented in #235.

PR #235, opened 2015-07-21
Two-way marshaling between NSData and ArrayBuffer

fealebenpae:
Introduce TNSDataAdapter to facilitate wrapping ArrayBuffers and ArrayBufferViews in an NSData interface without copying.
Introduce the interop.bufferFromData function which created an ArrayBuffer backed by an NSData instance.

Related to: https://github.com/NativeScript/ios-runtime/issues/231

### DIFF
--- a/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
+++ b/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
@@ -17,6 +17,7 @@
 
 #import "TNSArrayAdapter.h"
 #import "TNSDictionaryAdapter.h"
+#import "TNSDataAdapter.h"
 
 namespace NativeScript {
 using namespace JSC;
@@ -102,6 +103,8 @@ void ObjCConstructorBase::finishCreation(VM& vm, JSGlobalObject* globalObject, J
         this->_ffiTypeMethodTable.write = &writeAdapter<TNSArrayAdapter>;
     } else if (klass == [NSDictionary class]) {
         this->_ffiTypeMethodTable.write = &writeAdapter<TNSDictionaryAdapter>;
+    } else if (klass == [NSData class] || klass == [NSMutableData class]) {
+        this->_ffiTypeMethodTable.write = &writeAdapter<TNSDataAdapter>;
     } else {
         this->_ffiTypeMethodTable.write = &write;
     }
--- a/src/NativeScript/CMakeLists.txt
+++ b/src/NativeScript/CMakeLists.txt
@@ -61,6 +61,7 @@ set(HEADER_FILES
     ObjC/ObjCWrapperObject.h
     ObjC/TNSArrayAdapter.h
     ObjC/TNSDictionaryAdapter.h
+    ObjC/TNSDataAdapter.h
     Runtime/JSWeakRefConstructor.h
     Runtime/JSWeakRefInstance.h
     Runtime/JSWeakRefPrototype.h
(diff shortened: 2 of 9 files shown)


KristinaKoeva:
run ci

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
```diff
@@ -102,6 +103,8 @@ static void writeAdapter(ExecState* execState, const JSValue& value, void* buffe
         this->_ffiTypeMethodTable.write = &writeAdapter<TNSArrayAdapter>;
     } else if (klass == [NSDictionary class]) {
         this->_ffiTypeMethodTable.write = &writeAdapter<TNSDictionaryAdapter>;
+    } else if (klass == [NSData class]) {
```
jasssonpet: Maybe we should make this checks with `isSubclassOfClass` for the adapters, what do you think?

jasssonpet:
:+1:

jasssonpet:
I have a question about the wrapping and unwrapping of buffers:

In the following snippet:
```objective-c
var intRef = new interop.Reference(interop.types.int8, interop.alloc(4));
intRef[0] = 49;
intRef[1] = 50;
intRef[2] = 51;
intRef[3] = 52;
var data = NSData.dataWithBytesLength(intRef, 4);

var buffer = interop.bufferFromData(data);
var wrappedArrayBufferViewData = TNSObjCTypes.alloc().init().methodWithNSData(buffer);

console.log(wrappedArrayBufferViewData === data);
```

Should the result be `true` or `false`?

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
fealebenpae: I'll add a condition for `NSMutableData` as well, because `TNSDataAdapter` actually derives from `NSMutableData`. I don't think using `isSubclassOfClass` will work, though, because there are other descendants of `NSData`.

fealebenpae:
When you send an `ArrayBuffer` to something on the native side and we wrap it in `TNSDataAdapter` and then you get the adapter back, we'll unwrap the `ArrayBuffer` and hand it back to you, instead of an `NSData` object that is backed by the buffer. That's what we do for `TNSArrayAdapter` and `TNSDictionaryAdapter` as well.

However, when you have a native `NSData` object and wrap it in an `ArrayBuffer` with `bufferFromData`, passing the `NSData` object back and forth will always return the same object, not the wrapped buffer. In the sample case, `wrappedArrayBufferViewData === data.bytes` and `wrappedArrayBufferViewData === buffer` should be true.

Do you think we can improve this behavior in any way?

NathanaelA:
@fealebenpae do you know if this going to make it in 1.3? of the IOS runtimes?

fealebenpae:
This should make it in 1.3, yes. I haven't merged it yet because I'm waiting for some outside testing to see whether this actually satisfies a good number of use-cases. Can you share some details about your particular use-case?

NathanaelA:
@fealebenpae I am finishing up a WebSocket library for Android & IOS, and the IOS side uses a NSData decendant for the Binary side of things.     At the moment I created a "hack" in place that allows me to work around getting a NSData object.       At least on the IOS side once your marshalling makes it in; the hack will auto depreciate as the object will then be marshalled in as a ArrayBuffer rather than NSData.     On the Android side; I'm going to actually have to do a second release once the bug fix is committed as using ArrayBuffer's actually crash the Android runtime...

PR #247, opened 2015-07-28
Do not register the instance structure

KristinaKoeva:
When the inspector is enabled and a breakpoint is hit a Runtime.getDisplayableProperties message is dispatched. The logic behind Runtime.getDisplayableProperties enumerates getOwnPropertyNames and getOwnPropertySymbols and calls their toString method.
This results in a JSValue::toString call that defaults to JSCell::defaultValue which is guarded by a RELEASE_ASSERT_NOT_REACHED().

As a workaround do not register the instance structure

### DIFF
--- a/src/NativeScript/AllocatedPlaceholder.h
+++ b/src/NativeScript/AllocatedPlaceholder.h
@@ -30,8 +30,14 @@ public:
         return this->_wrappedObject;
     }
 
+    JSC::Structure* instanceStructure() const {
+        return this->_instanceStructure.get();
+    }
+
 private:
     id _wrappedObject;
+
+    JSC::WriteBarrier<JSC::Structure> _instanceStructure;
 
     AllocatedPlaceholder(JSC::VM& vm, JSC::Structure* structure)
         : Base(vm, structure) {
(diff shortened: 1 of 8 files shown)


src/NativeScript/AllocatedPlaceholder.h
```diff
@@ -40,8 +46,10 @@ class AllocatedPlaceholder : public JSC::JSDestructibleObject {
     void finishCreation(JSC::VM& vm, GlobalObject* globalObject, id wrappedObject, JSC::Structure* instanceStructure) {
         Base::finishCreation(vm);
         this->_wrappedObject = wrappedObject;
-        this->putDirect(vm, globalObject->instanceStructureIdentifier(), instanceStructure, JSC::ReadOnly | JSC::DontEnum | JSC::DontDelete);
+        _instanceStructure.set(vm, this, instanceStructure);
```
fealebenpae: Please prefix with `this->`.

src/NativeScript/ObjC/ObjCPrimitiveTypes.mm
```diff
@@ -23,13 +24,23 @@ static JSValue objCInstancetype_read(ExecState* execState, const void* buffer, J
...
-        });
+    Structure* structure;
+
+    if (ObjCConstructorBase* constructor = jsDynamicCast<ObjCConstructorBase*>(execState->thisValue())) {
+        structure = constructor->instancesStructure();
+    } else if (ObjCWrapperObject* wrapperObject = jsDynamicCast<ObjCWrapperObject*>(execState->thisValue())) {
+        structure = wrapperObject->structure();
+    } else if (ObjCSuperObject* superObject = jsDynamicCast<ObjCSuperObject*>(execState->thisValue())) {
+        structure = superObject->wrapperObject()->structure();
+    } else if (AllocatedPlaceholder* allocatedPlaceholder = jsDynamicCast<AllocatedPlaceholder*>(execState->thisValue())) {
```
fealebenpae: `AllocatedPlaceholder` is going to come up very often here. Please move this condition after the one for `ObjCConstructorBase`.

fealebenpae:
This looks good, I only have a couple of minor suggestions which I've expressed in inline comments. Also, could you please remove the `instanceStructureIdentifier` field and getter from `GlobalObject` seeing as how we don't need it anymore.

KristinaKoeva:
run ci

PR #248, opened 2015-07-30
Work around crash when logging warnings

fealebenpae:
Right now JSGlobalObjectConsoleClient::logWithLevel() can crash when there is no top call frame when called from NativeScript::warn


fealebenpae:
ping @jasssonpet

jasssonpet:
:+1:

PR #255, opened 2015-08-04
Timelines recording sends some long messages that hit the buffer limit of write method, so fallback to dispatch_io_write

KristinaKoeva:


KristinaKoeva:
run ci

PR #262, opened 2015-08-10
-[TNSValueWrapper value] can return null

fealebenpae:
When memory-managing an instance of a class derived from an Objective-C one there is that sliver of time after the JavaScript representation of the object has been garbage-collected, but the Objective-C representation is still alive. Calls to `retain` or `release` will try to access a JavaScript object that's not there anymore, so guard against that.


fealebenpae:
run ci

dtopuzov:
run ci

dtopuzov:
run ci

jasssonpet:
:+1:

PR #264, opened 2015-08-12
Call Block_copy in ObjCBlockCall::finishCreation

fealebenpae:
WTF::RetainPtr calls CFRetain on the block which is the same thing, but it ignores the return value which is crucial for blocks.


PanayotCankov:
:+1:

PR #272, opened 2015-08-18
Jasssonpet/require package json main suffix

### LINKED ISSUE #270: Better handling of package.json main configuration
Moved from https://github.com/NativeScript/nativescript-cli/issues/670

@jlooper and I have run across a few different npm modules that have a `"main"` of `"index"` in their `package.json`, instead of `"index.js"`—for example https://github.com/Sembiance/email-validator and https://github.com/snoj/email-validation.

This difference means that using one of these modules requires you to do:

``` js
var validator = require("email-validator/index");
```

instead of

``` js
var validator = require("email-validator");
```

Can the NativeScript CLI be changed to assume the .js suffix when it is not present? Furthermore, I'd like the CLI to support modules that do not have a `"main"` in their `package.json` at all. In these cases the NativeScript CLI should just use that module's `index.js` (if present).

Thanks.

jasssonpet:
Close #270

* Merge after: https://github.com/NativeScript/common-runtime-tests-app/pull/5

### DIFF
--- a/src/NativeScript/require.js
+++ b/src/NativeScript/require.js
@@ -73,7 +73,11 @@
                 //console.debug("PACKAGE_FOUND: " + packageJsonPath);
 
                 try {
-                    mainFileName = JSON.parse(packageJson).main || mainFileName;
+                    var packageJsonMain = JSON.parse(packageJson).main;
+                    if (packageJsonMain && !/\.js$/.test(packageJsonMain)) {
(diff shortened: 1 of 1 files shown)


jasssonpet:
run ci

jasssonpet:
run ci

jasssonpet:
run ci

fealebenpae:
:+1:

PR #274, opened 2015-08-20
Swift modules use mangled name which is different than the klassName so get the name from the metadata

KristinaKoeva:


src/NativeScript/TypeFactory.mm
```diff
@@ -271,11 +271,10 @@
     Structure* prototypeStructure = ObjCPrototype::createStructure(vm, globalObject, parentPrototype);
     ObjCPrototype* prototype = ObjCPrototype::create(vm, globalObject, prototypeStructure, metadata);
 
-    CString klassNameCString = klassName.utf8();
-    Class klass = objc_getClass(klassNameCString.data());
+    Class klass = objc_getClass(metadata->name());
     if (!klass) {
         SymbolLoader::instance().ensureFramework(metadata->topLevelModule()->getName());
-        klass = objc_getClass(klassNameCString.data());
+        klass = objc_getClass(klassName.utf8().data());
```
jasssonpet: I was just wondering if this should be `metadata->name()` also?

ivanbuhov: Yes, it should `metadata->name()`.

KristinaKoeva: will change it :disappointed:

jasssonpet:
:+1:

Issue #179, opened 2015-06-09
iOS 9 Support: Use a single metadata file


jasssonpet:
Research if there could be a way to ship only one metadata file for the right architecture.

PanayotCankov:
As suggested by @fealebenpae we can include the JavaScript files in the executable upon build. The app thinning will strip the fat binaries to a single slice on the AppStore when the app is distributed to the client devices. This will leave a single copy of the metadata in the actual download from the AppStore.

ivanbuhov:
Fixed https://github.com/NativeScript/ios-runtime/commit/1f9e76eac62e6f2a4cc21fcca6c1a9162aa40432

PR #275, opened 2015-08-20
Buhov/metadata merged in binary

ivanbuhov:
There is no more `metadata.bin` file. Metadata is now included in the executable upon build. The app thinning will strip the fat binaries to a single slice on the AppStore when the app is distributed to the client devices. This will leave a single copy of the metadata in the actual download from the AppStore.
Close https://github.com/NativeScript/ios-runtime/issues/179

### DIFF
--- a/cmake/GenerateMetadata.cmake
+++ b/cmake/GenerateMetadata.cmake
@@ -1,8 +1,11 @@
 macro(GenerateMetadata _target)
     add_dependencies(${_target} MetadataGenerator)
     add_custom_command(TARGET ${_target}
-        POST_BUILD
+        PRE_BUILD
         COMMAND "${CMAKE_SOURCE_DIR}/build/scripts/metadata-generation-build-step.sh"
         WORKING_DIRECTORY "${MetadataGenerator_BINARY_DIR}/bin"
+    )
+    target_link_libraries(${_target}
+    	"-sectcreate __DATA __TNSMetadata $(CONFIGURATION_BUILD_DIR)/metadata-$(CURRENT_ARCH).bin"
     )
 endmacro()
--- a/src/NativeScript/Metadata/Metadata.mm
+++ b/src/NativeScript/Metadata/Metadata.mm
@@ -207,8 +207,14 @@ void GlobalTable::iterator::findNext() {
     } while (this->_topLevelIndex < this->_globalTable->buckets.count);
 }
 
+static MetaFile* metaFileInstance(nullptr);
+
 MetaFile* MetaFile::instance() {
-    static MetaFile* instance(reinterpret_cast<MetaFile*>(loadFileInMemory((std::string("metadata-") + std::string(CURRENT_ARCH) + std::string(".bin")).c_str())));
-    return instance;
+    return metaFileInstance;
+}
+
+MetaFile* MetaFile::setInstance(void* metadataPtr) {
+    metaFileInstance = reinterpret_cast<MetaFile*>(metadataPtr);
+    return metaFileInstance;
 }
 }
--- a/build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
+++ b/build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
@@ -137,11 +137,11 @@
 			isa = PBXNativeTarget;
 			buildConfigurationList = 858B843018CA22B800AB12DE /* Build configuration list for PBXNativeTarget "__PROJECT_NAME__" */;
 			buildPhases = (
+				C97FD7AC1ADE5369004DB2A4 /* Generate Metadata */,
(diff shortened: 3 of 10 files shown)


src/NativeScript/Metadata/Metadata.mm
```diff
@@ -9,6 +9,7 @@
 #include <UIKit/UIKit.h>
 #include <sys/stat.h>
 #include "Metadata.h"
+#include <mach-o/getsect.h>
```
jasssonpet: is this header needed

ivanbuhov: Oh, it's no more needed. Forgot to remove it ;)

cmake/GenerateMetadata.cmake
```diff
@@ -1,8 +1,11 @@
 macro(GenerateMetadata _target)
     add_dependencies(${_target} MetadataGenerator)
     add_custom_command(TARGET ${_target}
-        POST_BUILD
+        PRE_BUILD
```
jasssonpet: Can this be made it in the `PRE_LINK` phase, so we don't generate metadata before each failed compile

ivanbuhov: You should be right, I will check it.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -207,8 +207,10 @@ int compareIdentifiers(const char* nullTerminated, const char* notNullTerminated
     } while (this->_topLevelIndex < this->_globalTable->buckets.count);
 }
 
+extern char startOfMetadataSection __asm("section$start$__DATA$__TNSMetadata");
```
fealebenpae: Does this work when the runtime is dynamically linked into the application? Also, we should try to stay away from ASM what with bitcode and everything. The mach-o APIs to resolve sections in the main binary didn't work out?

fealebenpae:
I can't tell from the patch, but did you move the 'Generate Metadata' build phase in the project template to the top? It would be best if it were right before the 'Link Binary' phase, for the same reason @jasssonpet cited for the proposed `PRE_LINK` CMake phase change.

ivanbuhov:
It turns out that the generation of `__TNSMetadata` section is part of the compilation phase, not the linking phase. It is the linker who does the job (`Ld` command is called), but it is called as part of the compilation process, right after all `CompileC` commands. So when the metadata generation step is between `Compile Sources` and `Link Binary With Libraries` or when it is set as `PRE_LINK` action the compilation phase fails with `ld: can't open -sectcreate file: /Users/buhov/Desktop/NS/ios-runtime/cmake-build/examples/Gameraww/Release-iphoneos/metadata-armv7.bin`, because it expects metadata to be already generated but there is no `metadata-armv7.bin` file yet. When I move the metadata generation step before `Compile Sources` phase (and in the other case - as `PRE_BUILD` action) it works like a charm. So I will leave it as it is unless you have better ideas.

src/NativeScript/Metadata/Metadata.mm
ivanbuhov: 1. I tested it with CLI project, so it works when the runtime is dynamically linked into the application. 
2. I had problems with Mach-O APIs. I didn't manage to retrieve a pointer to the Mach-O __DATA section. The documented way to achive it just didn't work, it returns me a wrong pointer pointing somewhere else but not to the section I am looking for. This is what i tried: 
```
const struct section_64* metadataSection = getsectbyname("__DATA", "__TNSMetadata");
static MetaFile* instance(reinterpret_cast<MetaFile*>(metadataSection->addr)); // note: the cast is working correctly, trust me :)
```
I will spend more time on this, maybe I miss something.

fealebenpae:
Alright, makes sense

src/NativeScript/Metadata/Metadata.mm
fealebenpae: A CLI project statically links the runtime, unless you modify it to link against the runtime framework. Please make sure that it all works when linked against the framework, even double-check if necessary.

Maybe we can merge it as-is, but then we should make migrating away from asm a medium-priority issue.

ivanbuhov: Ok, I will double check it.

I think I found a solution, [this guy](http://stackoverflow.com/questions/28978788/crash-reading-bytes-from-getsectbyname) has a similar problem and I will try his solution.

ivanbuhov:
Unfortunately, this approach is not working when NS Runtime is linked dynamically because it seems it is looking for TNSMetadata section in the dynamic library, but it is in the app executable.

jasssonpet:
Can we move the extern variable to the app then

ivanbuhov:
Yes, it should work. :+1:

fealebenpae:
:+1: after squashing to make things prettier

PR #276, opened 2015-08-21
Marshall NSDecimalNumber as Objective-C object wrapper - close #14

### LINKED ISSUE #14: NSDecimalNumber marshalling
Currently `NSDecimalNumber` (which inherits from `NSNumber`) is marshalled as JS number primitive and the precision is lost. We should create a wrapper object instead like we do for `NSString` and `NSMutableString`.

jasssonpet:


tests/TestRunner/app/Marshalling/ObjCTypesTests.js
```diff
@@ -204,6 +204,14 @@ describe(module.id, function () {
         expect(interop.handleof(buffer)).toBe(data.bytes);
     })
 
+    it("MethodWithNSDecimalNumber", function () {
+        expect(NSDecimalNumber.maximumDecimalNumber() instanceof NSDecimalNumber).toBe(true);
```
fealebenpae: Technically it'd be the same, but why'd you choose this over `expect(NSDecimalNumber.maximumDecimalNumber().toEqual(jasmine.any(NSDecimalNumber))`?

jasssonpet: I think that `instanceof` makes the code more clear.

fealebenpae: You've got me there, but wouldn't the error message would be more specific with the `toEqual` matcher if the expectation failed?

fealebenpae:
:+1:

PR #277, opened 2015-08-21
Refactor version tests - close #36

### LINKED ISSUE #36: Failing test on iOS 7.1.2 simulator
[This test](https://github.com/NativeScript/ios-runtime/blob/22864c5329d57b643d8a0ce2588fa4e32c5a55cd/tests/NativeScriptTests/NativeScriptTests/app/VersionDiffTests.js#L32) is failing on iOS 7.1.2 simulator.

The version diff tests should be rewritten not to depend on the iOS SDK, and should be more thorough testing all possible scenarios.

jasssonpet:


jasssonpet:
@ivanbuhov Do we store any availability metadat for enum members?

ivanbuhov:
@jasssonpet We store availability information in metadata for enums, but not for enum members. This can break us in some cases when enum member doesn't exist in older iOS version. Just adding availability information for enum members will not be enough to fix this, we have to make proxies for enums and implement `GetOwnPropertySlot` to check for availability for every enum member. What is your case?

fealebenpae:
Enums are compile-time constants. We can afford to always carry around enums for the latest iOS in the metadata, and access their values without conditions. The only place where we need strict availability information in the metadata is APIs we link to dynamically - classes, methods, functions, anything that's an exported symbol.

This means that structs, enums, their members, and any other kind of metadata member we don't rely on the environment for (either the Objective-C runtime, the CFBundle APIs, or `<dlfcn.h>`) can safely skip the availability check. Having them would be useless without the APIs to consume them, but accessing them won't be a runtime error, unlike accessing a class that isn't there.

ivanbuhov:
@fealebenpae consider the following imaginary case. In iOS 10 we have for example:
```
typedef NS_ENUM(NSInteger, NSComparisonResult) {
NSOrderedAscending = -1L, 
NSOrderedSame, 
NSOrderedDescending
};
```
In iOS 11 Apple introduce a new member:
```
typedef NS_ENUM(NSInteger, NSComparisonResult) {
NSOrderedAscending = -1L, 
NSOrderedSame, 
NSOrderedDescending,
NSOrderedRandom
};
```
In older iOS versions accessing `NSComparisonResult.NSOrderedRandom` will not break anything (it is just a constant) but `NSOrderedRandom` will appear in intellisense and it may break some APIs which expect `NSComparisonResult` and `NSOrderedRandom` is passed as argument. My question is: is this a case we should care for?

fealebenpae:
I don't think that should be a problem - adding a new enum member should preserve the values of the original members, that is to say, it should not be an ABI change. I'm not sure I've seen any enums or their members decorated with availability attributes, but even if there are, I can't think of a reason we should care about this, except in tooling scenarios where we could print warning during development when a constant value introduced in a later version of the OS is used.

jasssonpet:
> What is your case?

I just want to add tests for all supported scenarios atm.

fealebenpae:
:+1:

PR #284, opened 2015-08-28
ReferenceConstructor should accept a single type argument

fealebenpae:
and allocate an internal buffer of the correct size.

### DIFF
--- a/src/NativeScript/Marshalling/Reference/ReferenceConstructor.cpp
+++ b/src/NativeScript/Marshalling/Reference/ReferenceConstructor.cpp
@@ -9,6 +9,8 @@
 #include "ReferenceConstructor.h"
 #include "ReferencePrototype.h"
 #include "ReferenceInstance.h"
+#include "ReferenceTypeInstance.h"
+#include "RecordInstance.h"
 #include "Interop.h"
 
 namespace NativeScript {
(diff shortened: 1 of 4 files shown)


src/NativeScript/Marshalling/Reference/ReferenceConstructor.cpp
```diff
@@ -27,32 +27,32 @@ static EncodedJSValue JSC_HOST_CALL constructReference(ExecState* execState) {
...
-        }
-
-        bool hasHandle;
-        void* handle = tryHandleofValue(value, &hasHandle);
-        PointerInstance* pointer;
-        if (hasHandle) {
-            pointer = jsCast<PointerInstance*>(globalObject->interop()->pointerInstanceForPointer(execState->vm(), handle));
+        if (execState->argumentCount() == 2) {
+            JSValue value = execState->uncheckedArgument(1);
+            if (PointerInstance* pointer = jsDynamicCast<PointerInstance*>(value)) {
```
jasssonpet: What would happen if the value has a `handle`, but is not a `Pointer`. Do you think that this check could be made to `tryHandleofValue`?

fealebenpae: Well, `ObjCWrapperObject` and `RecordInstance` both have handles. We wouldn't want to dereference the handle of the former, though, so I didn't want to go with `tryHandleofValue`, although it makes sense to add a case where the value is a `RecordInstance` so that we can reuse its inner buffer. Save for records, can you think of any other cases where we'd want to reuse buffers?

jasssonpet: I think, that the problem with copying occurs with `Reference` also:

```
var point = new CGPoint();
var reference1 = new interop.Reference(CGPoint, point);
var reference2 = new interop.Reference(new interop.types.ReferenceType(CGPoint), reference1);

point.x = 2;

console.log(JSON.stringify(point)); // x: 2, y: 0
console.log(JSON.stringify(reference1[0])); // x: 0, y: 0
console.log(JSON.stringify(reference2[0][0])); // x: 0, y: 0
```

fealebenpae: So, `RecordInstance` and `ReferenceInstance` can be special-handled like `PointerInstance` is now?

jasssonpet: What's the problem using `handleOf`? We have a similar snippet in our docs:

```
var object = new UIView();

var ref = new interop.Reference(interop.types.id, object);
console.log(ref[0]); // isa: UIView
console.log(ref[1]); // _layer: CALayer
console.log(ref[2]); // _gestureInfo: null
```

fealebenpae: It dereferences the value - the snippet you linked to behaves like a pointer to the `NSObject` structure and would likely crash if the ivars weren't all ids already. I think the intention of `new interop.Reference(interop.types.id, object)` is such that `ref.value === object` ought to be true. What I'm trying to achieve is `NSObject **`, whereas the current effect is `*(NSObject *)`.

jasssonpet: Ok

fealebenpae:
run ci

fealebenpae:
@jasssonpet: I've updated the patch with handling for the cases we discussed, please take a look.

jasssonpet:
Looks good to me, :+1:

PR #285, opened 2015-08-31
Explicitly set the metadata generator's OS X deployment target

fealebenpae:
Otherwise it won't be able to run on an older OS X version than the one it was originally built on.


ivanbuhov:
:+1:

PR #297, opened 2015-09-11
Delay debugger pausing

KristinaKoeva:
If we pause right away the debugger messages that are send are not handled because the frontend is not yet initialized

### DIFF
--- a/src/debugging/TNSDebugging.h
+++ b/src/debugging/TNSDebugging.h
@@ -189,8 +189,18 @@ static void TNSEnableRemoteInspector(int argc, char **argv) {
 
       if (isWaitingForDebugger) {
         isWaitingForDebugger = NO;
-        CFRunLoopStop(CFRunLoopGetMain());
-        [inspector pause];
+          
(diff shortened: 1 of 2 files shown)


jasssonpet:
:+1:

src/debugging/TNSDebugging.h
```diff
@@ -189,8 +189,18 @@ static void TNSEnableRemoteInspector(int argc, char **argv) {
...
+        CFRunLoopPerformBlock(runloop,
+                              (__bridge CFTypeRef)(NSRunLoopCommonModes),
+                              ^{
+                                    // If we pause right away the debugger messages that are send are not handled because the frontend is not yet initialized
+                                    CFRunLoopRunInMode(kCFRunLoopDefaultMode, 1, false);
+                                    
+                                    [inspector pause];
+                               });
+        CFRunLoopWakeUp(runloop);
         CFRunLoopStop(CFRunLoopGetMain());
```
fealebenpae: You've already got the main runloop, you can use it here.

Issue #78, opened 2015-04-07
Uncaught error callback


jasssonpet:
The iOS runtime should invoke `global.__onUncaughtError(error)` (if set) on JavaScript error. After this function returns, the application will exit.

See also:
  * https://github.com/NativeScript/cross-platform-modules/issues/224
  * https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror
  * https://nodejs.org/api/process.html#process_event_uncaughtexception
  * https://docs.appcelerator.com/platform/latest/#!/api/Titanium.App-event-uncaughtException

PR #300, opened 2015-09-12
Call global.__onUncaughtError if a fatal exception is thrown

jasssonpet:
https://github.com/NativeScript/ios-runtime/issues/78

### DIFF
--- a/src/NativeScript/JSErrors.mm
+++ b/src/NativeScript/JSErrors.mm
@@ -13,6 +13,7 @@
 #import "TNSRuntime+Private.h"
 #import "TNSRuntime+Inspector.h"
 #import "TNSRuntime+Diagnostics.h"
+#import "JSWarnings.h"
 
 static TNSUncaughtErrorHandler uncaughtErrorHandler;
 void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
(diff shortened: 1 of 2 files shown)


src/NativeScript/JSErrors.mm
```diff
@@ -25,12 +25,23 @@ void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
...
-    globalObject->inspectorController().reportAPIException(execState, exception);
+
+    JSValue uncaughtError = globalObject->get(execState, Identifier::fromString(execState, "__onUncaughtError"));
+    if (uncaughtError.isFunction()) {
+        CallData uncaughtErrorFunctionCallData;
+        CallType uncaughtErrorFunctionCallType = getCallData(uncaughtError, uncaughtErrorFunctionCallData);
+
+        MarkedArgumentBuffer uncaughtErrorArguments;
+        uncaughtErrorArguments.append(exception->value());
+        call(execState, uncaughtError, uncaughtErrorFunctionCallType, uncaughtErrorFunctionCallData, jsUndefined(), uncaughtErrorArguments);
```
fealebenpae: Use the overload of `call` that excepts an error out parameter - right now if the uncaught error handler throws an error itself, this will reset the VM's exception field, which we will have cleared previously.

src/NativeScript/JSErrors.mm
```diff
@@ -25,12 +25,23 @@ void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
 
 void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception) {
     GlobalObject* globalObject = static_cast<GlobalObject*>(execState->lexicalGlobalObject());
-    globalObject->inspectorController().reportAPIException(execState, exception);
+
+    JSValue uncaughtError = globalObject->get(execState, Identifier::fromString(execState, "__onUncaughtError"));
+    if (uncaughtError.isFunction()) {
```
fealebenpae: Better replace this with a check that `getCallData` does not return `CallTypeNone` - `isFunction` checks whether a value is a cell that inherits from `JSFunction` or `InternalFunction`, but other cells (proxies, maybe?) can also override `getCallData`.

Also, `isFunction` will return true for ES6 constructor functions, but they aren't `[Call]`-able according to the spec.

jasssonpet:
@fealebenpae I've changed it so that if an error occurs in the `__onUncaughtError` callback, it will be logged to the terminal and the original error will be reported.

fealebenpae:
Lookin' good.

Issue #308, opened 2015-09-14
Move runtime files in project template


jasssonpet:
When the CLI updates a project, the following items should be easily replaced, so moving them in an internal folder will make this easier.

* `metadataGenerator`
* `NativeScript`
* `NativeScript.framework`
* `TNSDebugging.h`

jasssonpet:
@fealebenpae Should we also move `main.m` and `build.xcconfig`?

fealebenpae:
Well, `build.xcconfig` is supposed to be user-editable, so perhaps it would be best to leave it where it is so it doesn't get updated by the CLI.

I'm torn about `main.m`, though. My gut feeling says we should leave it where it is, but I'm afraid there will  come a time when we have to update it if we change the way the runtime is initialized (like scheduling it on a runloop, for example). @Fatme, your thoughts?

jasssonpet:
@fealebenpae What do you think if we move all NativeScript specific settings from the Xcode project to a separate nativescript-build.xcconfig?

For example:
* Link libraries (libNativeScript.a, libc++, libz, libicucore, MobileCoreServices)
* Other linker flags (-ObjC -sectcreate ...Metadata...)
* Header search paths (NativeScript/include)
* Library search paths (NativeScript/lib)

fealebenpae:
Sounds awesome. Would `build.xcconfig` include this new file?

jasssonpet:
We can also add three generic NativeScript build steps in the project template: `PRE_BUILD` `PRE_LINK` `POST_BUILD` to future proof any changes. We can also make the scripts language agnostic, by removing the `bash nativescript-pre-build.sh` and change it to `./nativescript-pre-build` for example.

fealebenpae:
+ aleph-one

jasssonpet:
There are some conflicts with our CocoaPods integration, that I'm currently looking at.

jasssonpet:
https://github.com/CocoaPods/CocoaPods/issues/1761#issuecomment-32890062

PR #314, opened 2015-09-16
Pass more Xcode build settings to the metadata generator

ivanbuhov:
Merge after https://github.com/NativeScript/ios-metadata-generator/pull/21


build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -209,7 +209,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "cd \"$SRCROOT/metadataGenerator/bin\"\nbash metadata-generation-build-step.sh";
+			shellScript = "cd \"$SRCROOT/metadataGenerator/bin\"\n./metadata-generation-build-step.py";
```
fealebenpae: I think we should remove the file extension from the script in order to future-proof it, in case we end up rewriting it in Ruby or something. Справка: https://github.com/NativeScript/ios-runtime/issues/308#issuecomment-140673487

PR #331, opened 2015-09-24
Fix cocoapods

jasssonpet:
* Moved the build.xcconfig to target level
* When installing cocoapods, it will not remove it, so we include cocoapods xcconfigs
* A not found xcconfig is not a build error

This is done because there is a Pod target that doesn't need NativeScript xcconfig.


ivanbuhov:
:+1:

jasssonpet:
I think that we should remove the xcconfig flag in the CLI build for the same reason:

```
-xcconfig filename
    Load the build settings defined in filename when building all targets. 
```

PR #336, opened 2015-09-25
Fix cocoapods xcconfig flags conflicts

jasssonpet:
https://github.com/CocoaPods/CocoaPods/issues/1761


jasssonpet:
This should be done for NativeScript plugins with xcconfig files, too.

Issue #215, opened 2015-07-07
Allow Blocking Calls to be Called Async


PanayotCankov:
There are some native APIs that execute as blocking calls. Currently the JavaScript execution has a locking mechanism implemented and we support JavaScript execution in multiple threads, however the JavaScript in these scenarios is serialized. Which means the two threads does not execute simultaneously, but instead the JavaScript parts wait for each-other. The native code however executes without locking the JS.

We should research if we can implement a function similar to "call" and "apply" on our methods that would run the native part of a blocking call on a second thread, and once complete, continue on the thread it was started.

For example if we can create "callAsync" method that returns a promise and consume it like:
``` JavaScript
var n = NSClass.alloc().init();

// Works for 5 seconds blocking the original thread and returns:
var result = n.getDataBlocking(5);
console.log(result);

// "callAsync" creates a promise, executes the getDataBlocking() on a background thread,
// the original thread continues execution, when complete "callAsync" returns to the
// original thread and resolves the promise, providing the result.
n.getDataBlocking.callAsync(n, 5).then(function(result) {
    // 5 sec. later this gets executed on the original thread. 
    console.log(result);
});
// The original thread continues immediately.
console.log("performing next statement");
```

This should be in no way substitute for multi-threading using WebWorkers.
> **WARNING:** This approach will degrade significantly if the native code actually performs chatty calls back to JavaScript!

PR #339, opened 2015-09-29
Add async method to FFICall

fealebenpae:
Implement #215

### DIFF
--- a/src/NativeScript/Calling/FFICall.cpp
+++ b/src/NativeScript/Calling/FFICall.cpp
@@ -8,14 +8,19 @@
 
 #include "FFICall.h"
 #include <malloc/malloc.h>
+#include <JavaScriptCore/JSPromiseDeferred.h>
+#include <JavaScriptCore/StrongInlines.h>
+#include <dispatch/dispatch.h>
 
 namespace NativeScript {
 using namespace JSC;
 
 const ClassInfo FFICall::s_info = { "FFICall", &Base::s_info, 0, CREATE_METHOD_TABLE(FFICall) };
 
-void FFICall::initializeFFI(VM& vm, JSCell* returnType, const Vector<JSCell*>& parameterTypes, size_t initialArgumentIndex) {
+void FFICall::initializeFFI(VM& vm, const InvocationHooks& hooks, JSCell* returnType, const Vector<JSCell*>& parameterTypes, size_t initialArgumentIndex) {
     ASSERT(this->methodTable()->destroy != FFICall::destroy);
+
+    this->_invocationHooks = hooks;
 
     this->_initialArgumentIndex = initialArgumentIndex;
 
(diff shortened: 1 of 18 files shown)


src/NativeScript/Runtime/ReleasePool.h
```diff
@@ -13,7 +13,7 @@
 #include <typeindex>
 #include <vector>
 #include <map>
-#include <stack>
+#include <wtf/Deque.h>
```
jasssonpet: I'm wondering - why did you use `Deque` here?

fealebenpae: I couldn't move items out of `std::stack`. `WTF::Deque` has exactly that in the `takeFirst()`/`takeLast()` methods.

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -72,24 +76,101 @@ void FFICall::visitChildren(JSCell* cell, SlotVisitor& visitor) {
...
+
+        if (InvocationHook post = callee->_invocationHooks.post) {
+            post(callee, execState, invocation);
         }
+
+        return JSValue::encode(result);
+    }
+
+    JSObject* FFICall::async(ExecState* execState, JSValue thisValue, const ArgList& arguments) {
+        Invocation* invocation = new Invocation(*this);
```
jasssonpet: I would suggest that you use `__block std::unique_ptr<Invocation>` so that the value could be properly released if we return from line 128 for example.

src/NativeScript/Runtime/ReleasePool.h
jasssonpet: I don't see using `first()` in your code, so shouldn't `pop()`/`top()` suffice.

src/NativeScript/Calling/FFICallPrototype.cpp
```diff
@@ -0,0 +1,51 @@
...
+        auto call = jsCast<FFICall*>(execState->thisValue());
+
+        JSValue array = execState->argument(1);
+
+        MarkedArgumentBuffer applyArgs;
+        if (!array.isUndefinedOrNull()) {
+            if (!array.isObject())
+                return throwVMTypeError(execState);
+            if (isJSArray(array)) {
+                if (asArray(array)->length() > maxArguments)
```
jasssonpet: Where does this `maxArguments` come from?

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -72,24 +76,101 @@ void FFICall::visitChildren(JSCell* cell, SlotVisitor& visitor) {
...
+            if (fakeExecState->hadException()) {
+                return fakeExecState->exception();
+            }
+        }
+
+        JSPromiseDeferred* deferred = JSPromiseDeferred::create(execState, execState->lexicalGlobalObject());
+        auto* releasePool = new ReleasePoolBase::Item(std::move(releasePoolHolder.relinquish()));
+
+        dispatch_async(dispatch_get_global_queue(0, 0), ^{
+            ffi_call(this->_cif, FFI_FN(invocation->function), invocation->_buffer + this->_returnOffset, reinterpret_cast<void**>(invocation->_buffer + this->_argsArrayOffset));
```
jasssonpet: I have a question. Could the `this` value be moved inside here. Also you could use the `invocation.resultBuffer()` method here.

jasssonpet:
:+1: after indentation.

I quite like the refactoring that you have done - makes the ffi calls more decoupled.

src/NativeScript/Runtime/ReleasePool.h
fealebenpae: For the love of me I couldn't make it to work with an `std::stack`, even with a liberal sprinkling of `std::move`. We can team up for a `std::stack` based implementation, but I'm not sure the effort is worth it.

src/NativeScript/Calling/FFICallPrototype.cpp
fealebenpae: I literally copied this from the old implementation of [`JSC::FunctionPrototype::functionProtoFuncApply`](https://github.com/WebKit/webkit/blob/fb7eff208db0fa9af1a929c6b6e26a32310caca3/Source/JavaScriptCore/runtime/FunctionPrototype.cpp#L127). `maxArguments` is a const declared in `Interpreter.h`.

src/NativeScript/Calling/FFICall.cpp
fealebenpae: Technically it's possible. I'll see if I can wrap it in a handle.

fealebenpae:
@jasssonpet I've tried to address the issues you pointed out. Please take a look and merge if you think the changes are ok.

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -72,24 +76,101 @@ void FFICall::visitChildren(JSCell* cell, SlotVisitor& visitor) {
...
+                CallData resolveCallData;
+                CallType resolveCallType = JSC::getCallData(deferred->resolve(), resolveCallData);
+
+                MarkedArgumentBuffer resolveArguments;
+                resolveArguments.append(result);
+                JSC::call(fakeExecState->lexicalGlobalObject()->globalExec(), deferred->resolve(), resolveCallType, resolveCallData, jsUndefined(), resolveArguments);
+            }
+
+            delete fakeCallFrame;
+            delete releasePool;
```
jasssonpet: I think that the `invocation` should be `.released()` here.

fealebenpae: Shouldn't the `unique_ptr` be destructed when the block is released?

PR #349, opened 2015-10-05
Throw error if there is a metadata for a symbol, but it's not available at runtime

jasssonpet:
I want to merge this in the release branch, if possible. The QAs are fine with that.

### DIFF
--- a/src/NativeScript/GlobalObject.mm
+++ b/src/NativeScript/GlobalObject.mm
@@ -286,13 +286,17 @@ bool GlobalObject::getOwnPropertySlot(JSObject* object, ExecState* execState, Pr
     }
     }
 
-    if (symbolWrapper) {
-        object->putDirectWithoutTransition(vm, propertyName, symbolWrapper);
-        propertySlot.setValue(object, None, symbolWrapper);
+    if (!symbolWrapper) {
+        WTF::String errorMessage = WTF::String::format("Metadata for \"%s.%s\" found but symbol not available at runtime.",
+                                                       symbolMeta->topLevelModule()->getName(), symbolMeta->name(), symbolMeta->name());
+        throwVMError(execState, createReferenceError(execState, errorMessage));
+        propertySlot.setValue(object, None, jsUndefined());
         return true;
     }
 
-    return false;
+    object->putDirectWithoutTransition(vm, propertyName, symbolWrapper);
+    propertySlot.setValue(object, None, symbolWrapper);
+    return true;
 }
 
 #ifdef DEBUG
--- a/tests/TestFixtures/Api/TNSApi.h
+++ b/tests/TestFixtures/Api/TNSApi.h
@@ -24,6 +24,8 @@ enum {
 
 extern NSString* const TNSConstant;
 
+extern NSString* const TNSUnavailableConstant;
+
 void functionThrowsException();
 
 @interface TNSApi : NSObject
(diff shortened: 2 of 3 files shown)


src/NativeScript/GlobalObject.mm
```diff
@@ -292,6 +301,12 @@ static EncodedJSValue JSC_HOST_CALL collectGarbage(ExecState* execState) {
         return true;
     }
 
+    if (isRuntimeUnavailable) {
```
KristinaKoeva: Could you check if the symbolWrapper is available instead ?

jasssonpet: It should be possible.

src/NativeScript/GlobalObject.mm
```diff
@@ -286,13 +286,15 @@ static EncodedJSValue JSC_HOST_CALL collectGarbage(ExecState* execState) {
     }
     }
 
-    if (symbolWrapper) {
-        object->putDirectWithoutTransition(vm, propertyName, symbolWrapper);
-        propertySlot.setValue(object, None, symbolWrapper);
+    if (!symbolWrapper) {
+        throwTypeError(execState, WTF::String::format("Metadata for \"%s\" found, but symbol \"%s\" is not available at runtime.", symbolMeta->jsName(), symbolMeta->name()));
```
ivanbuhov: Is this going to break our [ApiIterator test](https://github.com/NativeScript/ios-runtime/blob/master/tests/TestRunner/app/ApiTests.js#L443) because of some symbols which doesn't exist at runtime but there is metadata for them? If yes, we can add try-catch block in [ApiIterator test](https://github.com/NativeScript/ios-runtime/blob/master/tests/TestRunner/app/ApiTests.js#L443).

jasssonpet: Yes, you are totally right.

ivanbuhov:
:+1: after build succeeds.

jasssonpet:
The build is failing with:

```
2015-10-05 04:27:08.640 xcodebuild[70584:1007] [MT] iPhoneSimulator: SimVerifier returned: Error Domain=NSPOSIXErrorDomain Code=53 "Simulator verification failed." UserInfo=0x7fde9d5046c0 {NSLocalizedFailureReason=A connection to the simulator verification service could not be established., NSLocalizedRecoverySuggestion=Ensure that Xcode.app is installed on a volume with ownership enabled., NSLocalizedDescription=Simulator verification failed.}
```


which is irrelevant to this PR.

fealebenpae:
Why did you choose to throw `TypeError`? I think `ReferenceError` is more appropriate, although to be frank I'm not sure there is a widespread practice of `instanceof`-ing error objects anyway.

This change won't affect cases when we know the symbol shouldn't be there because of availability, right?

ivanbuhov:
@fealebenpae This change won't affect symbols which exist in metadata but are not exposed because of availability. These symbols hiding is handled on metadata API level, so [`Metadata::MetaFile::instance()->globalTable()->findMeta(symbolName)`](https://github.com/NativeScript/ios-runtime/blob/master/src/NativeScript/GlobalObject.mm#L210) will never return such symbol.

jasssonpet:
@fealebenpae Changed it to `ReferenceError`. I was too used to [TypeLoadException](https://msdn.microsoft.com/en-us/library/system.typeloadexception(v=vs.110).aspx) :D

PR #370, opened 2015-10-13
Fix out struct marshalling

jasssonpet:


src/NativeScript/Marshalling/Record/RecordConstructor.cpp
```diff
@@ -108,22 +108,25 @@ void RecordConstructor::write(ExecState* execState, const JSValue& value, void*
...
+
+                    field->ffiTypeMethodTable().write(execState, fieldValue, reinterpret_cast<void*>(reinterpret_cast<char*>(buffer) + field->offset()), field->fieldType());
+                    if (execState->hadException()) {
+                        return;
+                    }
                 }
             }
+        } else if (!value.isUndefinedOrNull()) {
+            JSValue exception = createError(execState, WTF::String::format("Could not marshall \"%s\" to \"%s\" type.", value.toWTFString(execState).utf8().data(), constructor->name(execState).utf8().data()));
+            reportFatalErrorBeforeShutdown(execState, Exception::create(execState->vm(), exception));
```
fealebenpae: You don't need to fatally shutdown here, you can just throw an exception in the ExecState. That way the error can by try/caught.

jasssonpet: @fealebenpae I've changed it to a JS error.

fealebenpae:
Is there a test for the new error-throwing behavior?

jasssonpet:
No, should I update the docs, too?

fealebenpae:
I don't think so, my opinion is that the new behavior is expected and intuitive. Let's just add a test so that we don't bork the error throwing.

jasssonpet:
@fealebenpae I believe this PR can be merged now.

PR #374, opened 2015-10-15
Persist NativeScript Inspector breakpoints - close #360

### LINKED ISSUE #360: Debugger - load the breakpoints from the previous session
It would be great to load the breakpoints from the previous debugging session.

jasssonpet:


src/debugging/Inspector/Inspector/NativeScript Inspector/ViewController.m
```diff
@@ -39,25 +48,42 @@ - (void)applicationShouldReopen:(NSNotification*)notification {
 - (void)setupWebView {
     WebView* webView = (WebView*)self.view;
 
-    JSContext* context = webView.mainFrame.javaScriptContext;
+    // Breakpoints are saved in window.localStorage so make them sandboxed per application
+    NSString* applicationSupportPath = NSSearchPathForDirectoriesInDomains(NSApplicationSupportDirectory, NSUserDomainMask, YES)[0];
+    NSString* appName = [[NSBundle mainBundle].infoDictionary objectForKey:(NSString*)kCFBundleNameKey];
+    NSString* localStoragePath = [NSString pathWithComponents:@[applicationSupportPath, appName, @"Local Storage", self->_projectName]];
```
KristinaKoeva: Shouldn't you update the ```self->_projectName``` when the application is reopened ?

jasssonpet: You are correct.

src/debugging/Inspector/Inspector/NativeScript Inspector/ViewController.m
```diff
@@ -2,26 +2,31 @@
 @import AppKit;
 @import WebKit;
 @import JavaScriptCore;
+#import <WebKit/WebPreferencesPrivate.h>
 #import "InspectorFrontendHost.h"
 
-@implementation ViewController
+@implementation ViewController {
+    NSString* _mainFileName;
```
KristinaKoeva: IMO if we have this as fields, we don't need to have parameterized methods ( ```update``` could just use ```self->_projectName``` ) but it's not that important

KristinaKoeva:
Looks great to me :+1:

PR #376, opened 2015-10-19
Remove extra space in console log output

jasssonpet:


src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
```diff
@@ -76,7 +76,9 @@ void GlobalObjectConsoleClient::printConsoleMessageWithArguments(MessageSource s
     } else {
         for (size_t i = 0; i < arguments->argumentCount(); ++i) {
             String argAsString = arguments->argumentAt(i).toString(arguments->globalState());
-            builder.append(' ');
```
KristinaKoeva: If I get this correctly, this issue arises when go through this if https://github.com/NativeScript/ios-runtime/pull/376/files#diff-5007059b61cbe1e81971cc464e602938R68
So could we just remove the space in ```": "``` ?

Issue #371, opened 2015-10-14
NativeScript apps crash on iPhone 6s


maknz:
We've tested this with our own app, and with a brand new {N} application. The app will crash when you try and run it, with some random memory error:

<img width="746" alt="screen shot 2015-10-14 at 5 43 07 pm" src="https://cloud.githubusercontent.com/assets/992356/10475060/5f8edc9a-729b-11e5-8332-58995fe1fada.png">

It *does* work fine with an iPhone 6s simulator, but **not** on a real iPhone 6s device.

fealebenpae:
Is there anything in the app's standard output?

maknz:
Doesn't seem to be. It has the output that the app usually has, up until it stops and highlights that line.

fealebenpae:
Alright, I'll try to reproduce it locally. I guess you're using CLI version 1.4.1, which is the latest on npm, the runtime should be 1.4.0, which is also the latest on npm, and Xcode 7.0.1, right? Are you running OS X Yosemite or El Capitan?

maknz:
Yes - 1.4.1 (latest) CLI, Xcode 7.0.1 from Mac App Store, OS X Yosemite.

If it's helpful, this crash occurred in the 1.3.0 environment (CLI and runtimes). We were, however, using an updated version of the core modules (master at the time, which became 1.4.0) as there were bugfixes in there that we needed.

fealebenpae:
Oh. Does the crash occur with the 1.3.0 modules and runtime? Also, can you give me the git commit SHA of the modules version you're using?

maknz:
I haven't tried with the 1.3.0 modules (haven't currently got the project in front of me, sorry). When we were using the 1.3.0 CLI and runtimes, we were using, at the time the crash was present on the 6s, the modules at https://github.com/NativeScript/NativeScript/commits/a396fd397ef71984a145f4c96256f6accacd9f12.

fealebenpae:
Unfortunately I don't have access to a 6s device yet, so I'm afraid I'll need to do some blind debugging.

I'm trying to determine whether I should look into the modules or the runtime for the reason behind the crash. To that end, I need to ask you to try to run a simple app sans modules on the phone: https://gist.github.com/fealebenpae/b48f18f2c40e0c1d3fd3.

maknz:
[...]

PR #377, opened 2015-10-19
Change arm64 part of libffi with a custom build

KristinaKoeva:
Our page alignment was causing problems (https://github.com/atgreen/libffi/blob/master/src/aarch64/sysv.S#L359) so change it with the value that is used by the objc runtime (http://www.opensource.apple.com/source/objc4/objc4-647/runtime/a1a2-blocktramps-arm64.s). @fealebenpae will apply the change in our libffi fork


KristinaKoeva:
Fixes https://github.com/NativeScript/ios-runtime/issues/371

PR #384, opened 2015-10-23
Make build steps to fail if an internal command fails

ivanbuhov:
Make build steps to fail if an internal command fails. Print meaningful message when metadata generation step fails.

### DIFF
--- a/build/project-template/internal/nativescript-post-build
+++ b/build/project-template/internal/nativescript-post-build
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -e
 
 pushd "$SRCROOT/internal"
 ./strip-dynamic-framework-architectures.sh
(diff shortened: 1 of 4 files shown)


build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -183,7 +183,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
-			shellScript = "\"$SRCROOT/internal/nativescript-pre-build\"";
+			shellScript = "set -e\n\"$SRCROOT/internal/nativescript-pre-build\"";
```
jasssonpet: I would advice against modifying the project template, if possible.

ivanbuhov: updated (y)

Issue #13, opened 2015-03-06
CFTypeRef marshalling from create or copy function


jasssonpet:
Currently `CFUUIDCreateString` returns a JS string primitive, which can't be `CFRelease`-d.

More info:
* [Create rule](https://developer.apple.com/library/mac/documentation/CoreFoundation/Conceptual/CFMemoryMgmt/Concepts/Ownership.html#//apple_ref/doc/uid/20001148-103029)
* [`CFUUIDCreateString`](https://developer.apple.com/library/mac/documentation/CoreFoundation/Reference/CFUUIDRef/index.html#//apple_ref/c/func/CFUUIDCreateString)

This should also be applicable to `CFDateRef` and `CFNumberRef`.

jasssonpet:
So there are still some functions like `CGPathCreateMutable()` which are not memory managed by the runtime. I'll have a look at them.

jasssonpet:
So it turns out that Swift has hardcoded the following types to be bridged: https://gist.github.com/anonymous/02bdf73ede2f68a6432a

jasssonpet:
We will memory manage only functions in `CF_IMPLICIT_BRIDGING_ENABLED` blocks. I'll modify the metadata generator to respect it.

jasssonpet:
Here is an update with other functions removed: https://gist.github.com/jasssonpet/327862790a058523179c/revisions

fealebenpae:
What's wrong with `UTTypeCopyDescription`? It's memory-managed even though it's not in an `CF_IMPLICIT_BRIDGING_ENABLED` block?

jasssonpet:
`UTTypeCopyDescription` returns JavaScript string which can't be released in NativeScript.

fealebenpae:
I'll create the `JSString` subclass that was supposed to manage primitives in the first place. However, do you think we might have to deal with `CFNumbeRef` values? The subclass-the-primitive approach won't work for numbers.

jasssonpet:
Well, there is `CFNumberCreate()`.

fealebenpae:
I meant to say `CFNumberRef`s returned from functions that would require a manual `CFRelease`.

jasssonpet:
All functions from the list above return `CFStringRef` only, so we shouldn't worry for `CFNumberRef` and `CFDateRef`, imo.

fealebenpae:
[...]

PR #386, opened 2015-10-27
Return an Unmanaged Type when we don't know how to memory manage an item

KristinaKoeva:
#13

https://github.com/NativeScript/ios-metadata-generator/pull/32

### DIFF
--- a/src/NativeScript/CMakeLists.txt
+++ b/src/NativeScript/CMakeLists.txt
@@ -2,9 +2,9 @@ set(HEADER_FILES
     NativeScript-Prefix.h
 
     Calling/FFICall.h
-    Calling/FFICallPrototype.h
     Calling/FFICallback.h
     Calling/FFICallbackInlines.h
+    Calling/FFICallPrototype.h
     Calling/FFIFunctionCall.h
     Calling/FFIFunctionCallback.h
     GlobalObject.h
@@ -71,6 +71,9 @@ set(HEADER_FILES
     ObjC/TNSArrayAdapter.h
     ObjC/TNSDataAdapter.h
     ObjC/TNSDictionaryAdapter.h
+    ObjC/Unmanaged/UnmanagedInstance.h
+    ObjC/Unmanaged/UnmanagedPrototype.h
+    ObjC/Unmanaged/UnmanagedType.h
     Runtime/JSWeakRefConstructor.h
     Runtime/JSWeakRefInstance.h
     Runtime/JSWeakRefPrototype.h
@@ -151,6 +154,9 @@ set(SOURCE_FILES
     ObjC/TNSArrayAdapter.mm
     ObjC/TNSDataAdapter.mm
     ObjC/TNSDictionaryAdapter.mm
+    ObjC/Unmanaged/UnmanagedInstance.cpp
+    ObjC/Unmanaged/UnmanagedPrototype.mm
+    ObjC/Unmanaged/UnmanagedType.cpp
     Runtime/JSWeakRefConstructor.cpp
     Runtime/JSWeakRefInstance.cpp
     Runtime/JSWeakRefPrototype.cpp
--- a/src/NativeScript/GlobalObject.h
+++ b/src/NativeScript/GlobalObject.h
@@ -109,6 +109,10 @@ public:
         return this->_weakRefInstanceStructure.get();
     }
 
+    JSC::Structure* unmanagedInstanceStructure() const {
+        return this->_unmanagedInstanceStructure.get();
+    }
+
     JSC::JSFunction* typeScriptOriginalExtendsFunction() const {
         return this->_typeScriptOriginalExtendsFunction.get();
     }
@@ -183,6 +187,8 @@ private:
     JSC::WriteBarrier<JSC::Structure> _weakRefConstructorStructure;
     JSC::WriteBarrier<JSC::Structure> _weakRefPrototypeStructure;
     JSC::WriteBarrier<JSC::Structure> _weakRefInstanceStructure;
+
+    JSC::WriteBarrier<JSC::Structure> _unmanagedInstanceStructure;
 
     std::map<Class, JSC::Strong<ObjCConstructorBase>> _objCConstructors;
 
--- a/src/NativeScript/ObjC/Unmanaged/UnmanagedInstance.h
+++ b/src/NativeScript/ObjC/Unmanaged/UnmanagedInstance.h
@@ -1,0 +1,49 @@
+#ifndef UnmanagedInstance_h
+#define UnmanagedInstance_h
+
+namespace NativeScript {
+class UnmanagedInstance : public JSC::JSDestructibleObject {
+public:
+    typedef JSC::JSDestructibleObject Base;
+
+    DECLARE_INFO;
+
+    static UnmanagedInstance* create(JSC::VM& vm, JSC::Structure* structure, JSC::JSCell* returnType, void* value = nullptr) {
+        UnmanagedInstance* object = new (NotNull, JSC::allocateCell<UnmanagedInstance>(vm.heap)) UnmanagedInstance(vm, structure);
+        object->finishCreation(vm, returnType, value);
+        return object;
+    }
+
+    static JSC::Structure* createStructure(JSC::JSGlobalObject* globalObject, JSC::JSValue prototype) {
+        return JSC::Structure::create(globalObject->vm(), globalObject, prototype, JSC::TypeInfo(JSC::ObjectType, StructureFlags), info());
+    }
+
+    JSC::JSCell* returnType() const {
+        return this->_returnType.get();
+    }
+
+    CFTypeRef data() const {
+        return this->_data;
+    }
+    void setData(void* data) {
+        this->_data = data;
+    }
+
+private:
+    CFTypeRef _data;
+    JSC::WriteBarrier<JSC::JSCell> _returnType;
+
+    static void visitChildren(JSC::JSCell*, JSC::SlotVisitor&);
+
+    static void destroy(JSC::JSCell* cell) {
+        JSC::jsCast<UnmanagedInstance*>(cell)->~UnmanagedInstance();
+    }
+
+    UnmanagedInstance(JSC::VM& vm, JSC::Structure* structure)
+        : Base(vm, structure) {
+    }
+
+    void finishCreation(JSC::VM&, JSC::JSCell*, void*);
+};
+}
+#endif /* UnmanagedInstance_h */
--- a/src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.h
+++ b/src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.h
@@ -1,0 +1,29 @@
+#ifndef UnmanagedPrototype_h
+#define UnmanagedPrototype_h
+
+namespace NativeScript {
+class UnmanagedPrototype : public JSC::JSNonFinalObject {
+public:
+    typedef JSC::JSNonFinalObject Base;
+
+    static UnmanagedPrototype* create(JSC::VM& vm, JSC::JSGlobalObject* globalObject, JSC::Structure* structure) {
+        UnmanagedPrototype* prototype = new (NotNull, JSC::allocateCell<UnmanagedPrototype>(vm.heap)) UnmanagedPrototype(vm, structure);
+        prototype->finishCreation(vm, globalObject);
+        return prototype;
+    }
+
+    DECLARE_INFO;
+
+    static JSC::Structure* createStructure(JSC::VM& vm, JSC::JSGlobalObject* globalObject, JSC::JSValue prototype) {
+        return JSC::Structure::create(vm, globalObject, prototype, JSC::TypeInfo(JSC::ObjectType, StructureFlags), info());
+    }
+
+private:
+    UnmanagedPrototype(JSC::VM& vm, JSC::Structure* structure)
+        : Base(vm, structure) {
+    }
+
+    void finishCreation(JSC::VM&, JSC::JSGlobalObject*);
+};
+}
+#endif /* UnmanagedPrototype_h */
--- a/src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
+++ b/src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
@@ -1,0 +1,46 @@
+#include "UnmanagedPrototype.h"
+#include "UnmanagedInstance.h"
+#include "ObjCTypes.h"
+#include "ObjCConstructorBase.h"
+
+namespace NativeScript {
+static char consumedUnmanagedCheck = 'k';
+
+using namespace JSC;
+
+const ClassInfo UnmanagedPrototype::s_info = { "Unmanaged", &Base::s_info, 0, CREATE_METHOD_TABLE(UnmanagedPrototype) };
+
+static EncodedJSValue takeValue(ExecState* execState, bool retained) {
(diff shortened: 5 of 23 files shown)


tests/TestRunner/app/ApiTests.js
```diff
@@ -465,6 +465,24 @@ describe(module.id, function () {
         expect(TNSReturnsRetained.newNSObjectMethod().retainCount()).toBe(1);
     });
 
+    it("unmanaged", function() {
```
jasssonpet: Formatting :)

tests/TestRunner/app/ApiTests.js
```diff
@@ -465,6 +465,24 @@ describe(module.id, function () {
...
+      expect('takeUnretainedValue' in unmanaged).toBe(true);
+      expect(functionReturnsUnmanaged().takeRetainedValue().retainCount()).toBe(1);
+
+      var value = functionReturnsUnmanaged().takeUnretainedValue();
+      expect(value.retainCount()).toBe(2);
+      CFRelease(value);
+
+      expect(function() {
+        (function() {
+          unmanaged.takeRetainedValue();
```
jasssonpet: The first call may be called before the `expect.toThrow` block, so that it doesn't eventually swallow an error if the first call throws. Also, what is the use of the self-executing function inside?

src/NativeScript/CMakeLists.txt
```diff
@@ -58,6 +58,9 @@ set(HEADER_FILES
     ObjC/Inheritance/ObjCClassBuilder.h
     ObjC/Inheritance/ObjCExtend.h
     ObjC/Inheritance/ObjCTypeScriptExtend.h
+    ObjC/Unmanaged/UnmanagedType.h
```
jasssonpet: I know it isn't obvious, but we try to keep these lists sorted, so that merge conflicts wouldn't happen as often if two or more users add files. If you wish to change it, the correct order here should be:

```
ObjC/TNSDictionaryAdapter.h
ObjC/Unmanaged/UnmanagedInstance.h
ObjC/Unmanaged/UnmanagedPrototype.h
ObjC/Unmanaged/UnmanagedType.h
Runtime/JSWeakRefConstructor.h
```

src/NativeScript/GlobalObject.h
```diff
@@ -22,6 +22,7 @@ class TypeFactory;
 class ObjCWrapperObject;
 class GlobalObjectInspectorController;
 class FFICallPrototype;
+class UnmanagedPrototype;
```
jasssonpet: I don't see this to be used in the header, maybe it should be removed?

src/NativeScript/ObjC/Unmanaged/UnmanagedType.h
```diff
@@ -0,0 +1,54 @@
+#ifndef UnmanagedType_h
+#define UnmanagedType_h
+
+#include <stdio.h>
+#include "FFIType.h"
+#include "UnmanagedPrototype.h"
+#include "UnmanagedInstance.h"
+
+#endif /* UnmanagedType_h */
```
jasssonpet: Shouldn't the `#endif` clause be at the end of the file?

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,51 @@
...
+
+void UnmanagedConstructor::visitChildren(JSCell* cell, SlotVisitor& visitor) {
+    Base::visitChildren(cell, visitor);
+
+    UnmanagedConstructor* constructor = jsCast<UnmanagedConstructor*>(cell);
+    visitor.append(&constructor->_returnType);
+}
+
+void UnmanagedConstructor::write(ExecState*, const JSValue&, void*, JSCell* self) {
+    ASSERT_NOT_REACHED();
```
jasssonpet: Maybe we can throw a JS Error if the user tries to pass an unmanaged object.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,51 @@
+#include "UnmanagedType.h"
+
+namespace NativeScript {
+using namespace JSC;
+
+const ClassInfo UnmanagedConstructor::s_info = { "Unmanaged", &Base::s_info, 0, CREATE_METHOD_TABLE(UnmanagedConstructor) };
+
+void UnmanagedConstructor::finishCreation(VM& vm, JSGlobalObject* globalObject, JSCell* returnType) {
```
jasssonpet: The `globalObject` parameter seems redundant.

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
```diff
@@ -0,0 +1,50 @@
...
+static char consumedUnmanagedCheck = 'k';
+
+using namespace JSC;
+
+const ClassInfo UnmanagedPrototype::s_info = { "Unmanaged", &Base::s_info, 0, CREATE_METHOD_TABLE(UnmanagedPrototype) };
+
+static EncodedJSValue JSC_HOST_CALL takeRetainedValue(ExecState* execState) {
+    UnmanagedInstance* instance = jsDynamicCast<UnmanagedInstance*>(execState->thisValue());
+    if (instance->data() == &consumedUnmanagedCheck) {
+        return throwVMTypeError(execState, WTF::ASCIILiteral("Unmanaged value has already been consumed. "));
```
jasssonpet: Is there for the trailing space here?

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    id result = static_cast<id>(instance->data());
+    instance->setData(&consumedUnmanagedCheck);
```
jasssonpet: I see the idea of the `consumedUnmanagedCheck`, but maybe a boolean field in the `UnmanagedInstance` be a clearer approach, imo.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.h
```diff
@@ -0,0 +1,54 @@
...
+    static JSC::Structure* createStructure(JSC::VM& vm, JSC::JSGlobalObject* globalObject, JSC::JSValue prototype) {
+        return JSC::Structure::create(vm, globalObject, prototype, JSC::TypeInfo(JSC::ObjectType, StructureFlags), info());
+    }
+
+    const FFITypeMethodTable& ffiTypeMethodTable() const {
+        return this->_ffiTypeMethodTable;
+    }
+
+private:
+    JSC::WriteBarrier<JSC::JSCell> _returnType;
```
jasssonpet: There isn't a specific order at the moment, but grouping fields together would make it easier for searching.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const void* data = *reinterpret_cast<void* const*>(buffer);
+
+    if (!data) {
+        return jsNull();
+    }
+
+    GlobalObject* globalObject = jsCast<GlobalObject*>(execState->lexicalGlobalObject());
+    UnmanagedConstructor* constructor = jsCast<UnmanagedConstructor*>(self);
+
+    return UnmanagedInstance::create(execState->vm(), UnmanagedInstance::createStructure(globalObject, globalObject->unmanagedInstanceStructure()->storedPrototype()), constructor->_returnType.get(), const_cast<void*>(data));
```
jasssonpet: I think that the idea of caching the structure in the global object is not to create one for each unmanaged instance, so `globalObject->unmanagedInstanceStructure()` should suffice for the second argument.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.h
```diff
@@ -0,0 +1,54 @@
...
+
+#include <stdio.h>
+#include "FFIType.h"
+#include "UnmanagedPrototype.h"
+#include "UnmanagedInstance.h"
+
+#endif /* UnmanagedType_h */
+
+namespace NativeScript {
+class UnmanagedConstructor : public JSC::JSDestructibleObject {
```
jasssonpet: Maybe the file should be renamed to be in sync with the class.

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
KristinaKoeva: That was my idea at first but then @fealebenpae suggested this approach as a cooler way to do it, so I decided to follow his advice :) Plus you don't need to add another member to the header. If we all agree on your suggestion, I will change it :)

src/NativeScript/ObjC/Unmanaged/UnmanagedInstance.h
```diff
@@ -0,0 +1,49 @@
...
+
+    void* data() const {
+        return this->_data;
+    }
+    void setData(void* data) {
+        this->_data = data;
+    }
+
+private:
+    void* _data;
```
fealebenpae: Let's make this `CFTypeRef` to make it explicit what we're dealing with.

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.h
```diff
@@ -0,0 +1,33 @@
+#ifndef UnmanagedPrototype_h
+#define UnmanagedPrototype_h
+
+namespace NativeScript {
+class UnmanagedPrototype : public JSC::JSDestructibleObject {
```
fealebenpae: `UnmanagedPrototype` doesn't need destruction, so it can be a plain old `JSNonFinalObject`.

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.h
```diff
@@ -0,0 +1,33 @@
...
+    static JSC::Structure* createStructure(JSC::VM& vm, JSC::JSGlobalObject* globalObject, JSC::JSValue prototype) {
+        return JSC::Structure::create(vm, globalObject, prototype, JSC::TypeInfo(JSC::ObjectType, StructureFlags), info());
+    }
+
+private:
+    UnmanagedPrototype(JSC::VM& vm, JSC::Structure* structure)
+        : Base(vm, structure) {
+    }
+
+    static void destroy(JSC::JSCell* cell) {
```
fealebenpae: See my comment above.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+JSValue UnmanagedType::read(ExecState* execState, const void* buffer, JSCell* self) {
+    const void* data = *reinterpret_cast<void* const*>(buffer);
+
+    if (!data) {
+        return jsNull();
+    }
+
+    GlobalObject* globalObject = jsCast<GlobalObject*>(execState->lexicalGlobalObject());
+    UnmanagedType* constructor = jsCast<UnmanagedType*>(self);
```
fealebenpae: Rename `constructor` to something else.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+    return UnmanagedInstance::create(execState->vm(), globalObject->unmanagedInstanceStructure(), constructor->_returnType.get(), const_cast<void*>(data));
+}
+
+void UnmanagedType::visitChildren(JSCell* cell, SlotVisitor& visitor) {
+    Base::visitChildren(cell, visitor);
+
+    UnmanagedType* constructor = jsCast<UnmanagedType*>(cell);
```
fealebenpae: Ditto.

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+void UnmanagedType::write(ExecState* execState, const JSValue&, void*, JSCell* self) {
+    throwVMTypeError(execState, WTF::ASCIILiteral("Unmanaged type could not be consumed from native. Try calling takeRetained/takeUnretained before."));
+}
+
+bool UnmanagedType::canConvert(ExecState*, const JSValue&, JSCell* self) {
+    return false;
+}
+
+const char* UnmanagedType::encode(JSCell* cell) {
```
fealebenpae: I think this can be `RELEASE_ASSERT_NOT_REACHED()`. It doesn't make sense for the Unmanaged type to be able to encode itself using Objective-C's method encodings.

KristinaKoeva: I agree, @jasssonpet doesn't so both of you make up you mind on whether you want 
1. Some implementation of those methods
2. No implementation of those methods and RELEASE_ASSERT_NOT_REACHED/ ASSERT_NOT_REACHED
3. Something else
:)

src/NativeScript/ObjC/Unmanaged/UnmanagedPrototype.mm
```diff
@@ -0,0 +1,50 @@
...
+
+    ObjCConstructorBase* returnType = jsCast<ObjCConstructorBase*>(instance->returnType());
+    JSValue retainedValue = NativeScript::toValue(execState, result, returnType->klass());
+
+    [result release];
+
+    return JSValue::encode(retainedValue);
+}
+
+static EncodedJSValue JSC_HOST_CALL takeUnretainedValue(ExecState* execState) {
```
ivanbuhov: `takeRetainedValue` and `takeUnretainedValue` methods differ in one line only (https://github.com/NativeScript/ios-runtime/pull/386/files#diff-d4cef6d3713c4250ebb83cd1a1d4ac25R25). What do you think about extracting the common logic in another method and calling it twice with different arguments? Something like this:
```
static EncodedJSValue takeValue(ExecState* execState, bool retained) {
    ...
    if(retained) {
        [result release];
    }
    ...
}

static EncodedJSValue JSC_HOST_CALL takeRetainedValue(ExecState* execState) {
    return takeValue(execState, true);
}

static EncodedJSValue JSC_HOST_CALL takeUnretainedValue(ExecState* execState) {
    return takeValue(execState, false);
}
```

KristinaKoeva: Will do :)

src/NativeScript/ObjC/Unmanaged/UnmanagedType.cpp
fealebenpae: @jasssonpet, what do you think?

jasssonpet: Although not used at the moment, I think that it's best to leave the `encode` implementation. (Instead of returning an empty string, we can have `RELEASE_ASSERT_NOT_REACHED()`.)

fealebenpae: In what scenario would we need `UnamangedType` to be able to encode itself?

jasssonpet: As you know, the `encode` method is used only for computing the type encoding when extending an Objective-C class. I agree, that it's sound unreasonable now, but my feeling is that the `UnmanagedType` is able to generate it's encoding and is best to do so, because we don't have any information about where this type will be used here.

jasssonpet:
:+1:

ivanbuhov:
:+1:

PR #390, opened 2015-10-29
Remove LaunchScreen.xib from project.

e2l3n:
<b>Description</b>: Remove `LaunchScreen.xib` from project template.
<b>TeamPulse</b>: [item/296671](http://teampulse.telerik.com/view#item/296671)

<b>Notes</b>: The `UILaunchStoryboardName` property causes the application to ignore the `Default.png` and `Default@2x.png` launch images on 4-inch displays.
<i>Edit</i>: This is the ticket - [admin.telerik.com](https://admin.telerik.com/?Context=Support&Control=~/UserControls/Support/ThreadReply.ascx&ST=949265&SM=0&PURL=%2f%3fContext%3dSupport%26Control%3d%257e%252fUserControls%252fSupport%252fMySupportThreads.ascx%26A%3d0%26SearchType%3d0)

PR #394, opened 2015-11-02
Revert "Remove LaunchScreen.xib from project."

e2l3n:
Reverts NativeScript/ios-runtime#390

<b>Description</b>: Restore `LaunchScreen.xib` and `UILaunchStoryboardName` plist key due to scalling issues. I will continue investigating the issue mentioned in https://github.com/NativeScript/ios-runtime/pull/390.


fealebenpae:
@dtopuzov and I agreed that the correct way to handle this is to add the proper launch screen images to the hello-world template instead of applying this revert.

PR #397, opened 2015-11-04
Define UILaunchImages key in project plist.

e2l3n:
<b>Description</b>: This PR adds [`UILaunchImages`](https://developer.apple.com/library/ios/documentation/General/Reference/InfoPlistKeyReference/Articles/iPhoneOSKeys.html) key to the project plist. It replicates `Cordova`'s solution.
<b>Notes</b>: Testing showed that removing the `UILaunchStoryboardName` causes scaling issues on larger retina screens. Depending on correct default launch image names isn't enough to fix the scaling issue, it appears that the key `UILaunchImages` resolves the issue.

### DIFF
--- a/build/project-template/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
+++ b/build/project-template/__PROJECT_NAME__/__PROJECT_NAME__-Info.plist
@@ -41,5 +41,108 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>UILaunchImages</key>
+    <array>
+        <dict>
+            <key>UILaunchImageMinimumOSVersion</key>
+            <string>8.0</string>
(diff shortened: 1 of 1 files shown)


fealebenpae:
Way back when, in the dawn of NativeScript, the Xcode template for the iOS runtime actually included entries that described launch images and app icons. It turned out to be less than ideal, because hardcoding every possible launch screen combination actually precludes users from submitting non-universal apps (e.g. iPhone or iPad-only) to the App Store. Users had to manually edit the project under `platforms/ios` in order to get iTunes Connect to accept their app.

NativeScript apps are structured in such a way that manual changes to the platforms folders are highly discouraged and problematic - `platform remove`, `platform add` and friends can overwrite the changes to the platform folder.

A far better place to store the metadata about launch screens (and the launch screens themselves) is the designated `App_Resources` folder. I will not accept any changes into the Xcode template that would necessitate user intervention in the platform folder. Please find another way to resolve this - like https://github.com/NativeScript/nativescript-cli/issues/1089 for example.

PR #404, opened 2015-11-13
Update metadata generator

### LINKED ISSUE #373: Loading static frameworks as shared frameworks
We're trying to use Raygun4iOS, which is a crash reporting library. I've tried specifying it in a plugin's Podfile, and just putting the framework into platforms/ios in the plugin. Regardless, when the app is starting, we are told that NativeScript could not load the framework -- as a log line when the app is starting and listing the frameworks it's loading. There's no other error detail though, so I don't know why this would be happening.

Instructions for setting up Raygun in a regular iOS application are here: https://raygun.io/docs/languages/ios.

jasssonpet:
https://github.com/NativeScript/ios-metadata-generator/pull/35/


jasssonpet:
run ci

KristinaKoeva:
run ci

PR #408, opened 2015-11-18
Expose build configuration as an enviromental variable

jasssonpet:
One way to read it from JavaScript:

```javascript
NSProcessInfo.processInfo().environment.objectForKey('BUILD_CONFIGURATION'); // Debug, Release, ...
```

ping @PanayotCankov

### DIFF
--- a/build/project-template/internal/main.m
+++ b/build/project-template/internal/main.m
@@ -20,9 +20,11 @@ int main(int argc, char *argv[]) {
 #ifndef NDEBUG
     NSString *libraryPath = [NSSearchPathForDirectoriesInDomains(
         NSLibraryDirectory, NSUserDomainMask, YES) firstObject];
-    NSString *liveSyncPath =
-        [NSString pathWithComponents:
-                      @[ libraryPath, @"Application Support", @"LiveSync" ]];
+    NSString *liveSyncPath = [NSString pathWithComponents:@[
+      libraryPath,
+      @"Application Support",
+      @"LiveSync"
+    ]];
     NSString *appFolderPath =
         [NSString pathWithComponents:@[ liveSyncPath, @"app" ]];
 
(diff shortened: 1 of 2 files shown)


jasssonpet:
@PanayotCankov  It would be interesting if we could also reuse something like [envify](https://github.com/hughsk/envify) to completely strip debugging code during minification in release builds.

fealebenpae:
Why do we need this?

PanayotCankov:
@fealebenpae There is some extensive logging and debug information that we do not want to generate when the app is built for release.
@jasssonpet The suggestion looks cool, we'll have that in mind.

fealebenpae:
@PanayotCankov I'm hesitant to involve the runtime in this, especially since this adds further complexity to the Xcode project template. Have you considered a CLI-based solution instead?

PanayotCankov:
@fealebenpae Yes, I too share your concerns. If we introduce this in the CLI we will move the complexity from the runtime to the CLI. Then we will bind some of the debugging functionality to the CLI, rendering it inaccessible when using Xcode only from the Xcode project generated in the platforms/ios. Currently I can leave the debugging info enabled and worry about switching it on and off at a later stage, probably there are other solutions (like treating *.release.js and *.debug.js files in the CLI). If you think you'll be bothered too much supporting @jasssonpet's suggestion please feel free to reject the PR.

fealebenpae:
My chief concern is that we've introduced enough complexity in the template to handle the various quirks of the CLI, its plugin system, and CocoaPods to boot. Adding more on top of that would make it even harder to use the runtime in an existing application since all of the template's complexity would have to be added to the existing project.

Now, if the CLI supported adding raw source files as part of a plugin (similarly to Cordova plugins), `nativescript-core-modules` could conceivably include a dummy header file that exports the build configuration as a constant:
```
#ifndef NDEBUG
static const BOOL TNSCoreModulesDebug = YES;
#else
static const BOOL TNSCoreModulesDebug = NO;
#endif
```

PanayotCankov:
The CLI already treats the tns-core-modules as a plugin and I think we will be soon shipping the android "widgets" JAR that has the layout implementation as part of the modules. So shipping the header you mention in the tns-core-modules will actually be pretty neat. Do you want to discuss further what have to be done in the CLI in order to allow raw source files in the plugins?

fealebenpae:
Yes, let's loop @teobugslayer on this IRL. I'll close this for the time being.

PanayotCankov:
@fealebenpae :+1:

PR #410, opened 2015-11-18
Disable JIT

ivanbuhov:
We are disabling JIT because of a bug found in the latest webkit version, causing some tests to fail. We can't figure out the actual piece of code causing the bug, but disabling JIT makes it disappear.

P.S. Til now, JIT was enabled only on simulator, because of iOS restrictions, so the change will not affect device builds.

### DIFF
--- a/src/NativeScript/TNSRuntime.mm
+++ b/src/NativeScript/TNSRuntime.mm
@@ -33,6 +33,7 @@ using namespace NativeScript;
 + (void)initialize {
     if (self == [TNSRuntime self]) {
         initializeThreading();
+        JSC::Options::useJIT() = false;
     }
 }
 


src/NativeScript/TNSRuntime.mm
```diff
@@ -42,6 +42,7 @@ + (void)initializeMetadata:(void*)metadataPtr {
 
 - (instancetype)initWithApplicationPath:(NSString*)applicationPath {
     if (self = [super init]) {
+        JSC::Options::useJIT() = false;
```
jasssonpet: Maybe we should move this line of code to the static constructor of the runtime, if it's affecting all instances of the runtime.

fealebenpae: `+[TNSRuntime initialize]` is a good candidate.

PR #425, opened 2015-11-30
Unbox primitives when marshalling to id

fealebenpae:


src/NativeScript/ObjC/ObjCTypes.mm
```diff
@@ -118,7 +119,7 @@ id toObject(ExecState* execState, const JSValue& value) {
     }
 
     if (value.isString()) {
-        return [NSString stringWithString:(NSString*)value.toString(execState)->value(execState).createCFString().get()];
+        return [static_cast<NSString *>(jsCast<JSString*>(value)->value(execState)) copy];
```
jasssonpet: I wonder if we should `autorelease` the copy.

fealebenpae: Yes, we should, and I'll fix it right away.

jasssonpet:
:+1:

I wonder if it's worth adding the same logic when we marshall to types other than `id`.

fealebenpae:
We don't need to. Calling `JSValue::toNumber(ExecState*)` performs unboxing for us. Same with strings and booleans.

PR #426, opened 2015-12-01
Use JSAgentContext as all the other agents

KristinaKoeva:
We use to wait for didCreateFrontendAndBackend to get the frontend and backend in the domain agents but now a null pointers are passed so use a JSAgentContext instead.

### DIFF
--- a/src/NativeScript/inspector/GlobalObjectInspectorController.cpp
+++ b/src/NativeScript/inspector/GlobalObjectInspectorController.cpp
@@ -80,28 +80,18 @@ GlobalObjectInspectorController::GlobalObjectInspectorController(GlobalObject& g
     , m_frontendRouter(FrontendRouter::create())
     , m_backendDispatcher(BackendDispatcher::create(m_frontendRouter.copyRef()))
     , m_globalObject(globalObject)
+    , m_agentContext({ *this, *m_injectedScriptManager, m_frontendRouter.get(), m_backendDispatcher.get() })
+    , m_jsAgentContext(m_agentContext, m_globalObject)
+
 {
-    AgentContext baseContext = {
-        *this,
-        *m_injectedScriptManager,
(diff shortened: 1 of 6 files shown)


src/NativeScript/inspector/GlobalObjectInspectorController.cpp
```diff
@@ -125,7 +125,19 @@ GlobalObjectInspectorController::~GlobalObjectInspectorController() {
 }
 
 void GlobalObjectInspectorController::registerDomainDispatcher(WTF::String domainIdentifier, JSC::JSCell* constructorFunction) {
-    std::unique_ptr<DomainInspectorAgent> domainInspectorAgent = std::make_unique<DomainInspectorAgent>(m_globalObject, domainIdentifier, constructorFunction);
+    AgentContext baseContext = {
```
KristinaKoeva: I wanted to make that a class field but couldn't initialize it in the constructor. I am open for suggestions on how to do it :) 

```{ { *this, *m_injectedScriptManager, m_frontendRouter.get(), m_backendDispatcher.get() }, globalObject }``` fails with
```InspectorAgentBase.h:51:5: Candidate constructor not viable: cannot convert initializer list argument to 'Inspector::AgentContext &'```

jasssonpet: I have a suspicion that the reason you can't do it inline is because the  `AgentContext&` parameter isn't marked as `const` in the constructor: https://gist.github.com/anonymous/306cfb0e33be4c2e9855.

KristinaKoeva: Yeah, I've found a stackoverflow answer that said the same thing, but since it's not cool to modify the source I guess I am stuck with creating a context on every domain register

PR #421, opened 2015-11-25
Add inspector build script

jasssonpet:
@KristinaKoeva, could you give a hand with the code signing?

This should add no more than a minute to the build time (except the first time).

PR #436, opened 2015-12-09
Fix window dragging on El Capitan

KristinaKoeva:


KristinaKoeva:
Would somebody that hasn't upgraded to El Capitan verify whether this is working for older OS too :)

jasssonpet:
It's working for me (OS X Yosemite), but I think that the preprocessor check is always false, because our `MACOSX_DEPLOYMENT_TARGET` is 10.10.

KristinaKoeva:
You are completely right, I blindly copy pasted this check from webkit source :disappointed: I will remove it so that you could test again

jasssonpet:
The `startWindowDrag` is not being called, so I suspect everything is working :smile:

KristinaKoeva:
This one https://github.com/NativeScript/ios-runtime/blob/master/src/debugging/WebInspectorUI/Base/Main.js#L1562 takes care of it :smiley_cat:

jasssonpet:
BTW, it seems that #421 won't make it to the 1.5.1 milestone, so maybe you'll have to build the Inspector manually one last time.

PR #437, opened 2015-12-09
Add all properties of domain types to the TypeScript generated glue code

KristinaKoeva:
Also expose inspectorTimestamp method that would be used from modules guys to record a timeline event. Doing that in JavaScript was not possible because the elapsed time is based on some context that is held and modified by several different places


src/NativeScript/inspector/GlobalObjectInspectorController.cpp
```diff
@@ -74,6 +74,14 @@ JSC::EncodedJSValue registerDispatcher(JSC::ExecState* execState) {
     return JSValue::encode(jsUndefined());
 }
 
+JSC::EncodedJSValue inspectorTimestamp(JSC::ExecState* execState) {
```
jasssonpet: I wonder if this function should be marked with `JSC_HOST_CALL`?

PR #438, opened 2015-12-10
Performance improvements for method calls

fealebenpae:
This yields a ~100% performance improvement in synthetic benchmarks.

### DIFF
--- a/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
+++ b/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
@@ -21,6 +21,10 @@
 #include "Interop.h"
 #include "TNSFastEnumerationAdapter.h"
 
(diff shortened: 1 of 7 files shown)


src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
```diff
@@ -21,6 +21,10 @@
 #include "Interop.h"
 #include "TNSFastEnumerationAdapter.h"
 
+@protocol TNSDerivedClass
```
jasssonpet: Can we completely remove this protocol now?

fealebenpae: It's still used within the class builder and I can't figure out a substitution. It's not on a hot path there so I guess it's fine.

PR #440, opened 2015-12-11
Update changelog

KristinaKoeva:


CHANGELOG.md
```diff
@@ -1,3 +1,16 @@
+1.5.1
+=====
+
+## Bug Fixes
+- [Framework detection in metadata generator is not working for every case](https://github.com/NativeScript/ios-runtime/issues/420)
```
jasssonpet: This is actually a bug that was introduced in an attempt to fix: https://github.com/NativeScript/ios-runtime/issues/373, so maybe it'll be more clear to list that instead :smile_cat:

jasssonpet:
:+1:

PR #444, opened 2015-12-16
Replace ObjC association with WeakGCMap

fealebenpae:
This nets us a few small performance wins because we no longer use the thread-safe Objective-C association API to store JavaScript wrappers in native objects.

### DIFF
--- a/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
+++ b/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
@@ -16,6 +16,7 @@
 #include "ObjCMethodCallback.h"
 #include "ObjCSuperObject.h"
 #include "ObjCProtocolWrapper.h"
+#include "ObjCWrapperObject.h"
 #include "ObjCTypes.h"
 #include "FFIType.h"
 #include "Interop.h"
@@ -65,8 +66,7 @@ static void attachDerivedMachinery(GlobalObject* globalObject, Class newKlass, J
         Structure* superStructure = ObjCSuperObject::createStructure(vm, globalObject, superPrototype);
         ObjCSuperObject* superObject = ObjCSuperObject::create(vm, superStructure, derivedWrapper, globalObject);
         derivedWrapper->putDirect(vm, vm.propertyNames->superKeyword, superObject, ReadOnly | DontEnum | DontDelete);
-
-        [TNSValueWrapper attachValue:derivedWrapper toHost:instance];
+        
         return instance;
     });
     class_addMethod(metaClass, @selector(allocWithZone:), newAllocWithZone, "@@:");
@@ -74,11 +74,9 @@ static void attachDerivedMachinery(GlobalObject* globalObject, Class newKlass, J
     IMP retain = findNotOverridenMethod(newKlass, @selector(retain));
     IMP newRetain = imp_implementationWithBlock(^(id self) {
         if ([self retainCount] == 1) {
-            if (TNSValueWrapper* wrapper = objc_getAssociatedObject(self, globalObject->JSScope::vm())) {
-                if (JSObject* object = wrapper.value) {
-                    JSLockHolder lockHolder(globalObject->vm());
-                    gcProtect(object);
-                }
+            if (JSObject* object = globalObject->interop()->objectMap().get(self)) {
+                JSLockHolder lockHolder(globalObject->vm());
+                gcProtect(object);
             }
         }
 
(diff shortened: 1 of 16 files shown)


src/NativeScript/ObjC/TNSArrayAdapter.mm
```diff
@@ -73,4 +74,13 @@ - (NSUInteger)countByEnumeratingWithState:(NSFastEnumerationState*)state objects
     return count;
 }
 
+- (void)dealloc {
+    {
+        JSLockHolder lock(self->_execState);
```
ivanbuhov: Why do we need to lock here but not in `ObjCWrapperObject::~ObjCWrapperObject()`?

fealebenpae: `ObjCWrapperObject::~ObjCWrapperObject()` is a `JSCell` destructor and that is invoked during garbage collection where it is guaranteed that the API lock on the virtual machine is acquired.

ivanbuhov: I see ;)

jasssonpet:
The following snippet crashes on my iPhone 4s simulator:

```javascript
var obj = NSObject.new();
var arr = new NSArray([obj]);
console.log(arr[0]);
```

```
ASSERTION FAILED: m_ptr
/Users/jzhekov/Desktop/Projects/Kimera/ios-runtime/src/WebKit/Source/WTF/wtf/RefPtr.h(68) : T &WTF::RefPtr<JSC::JSLock>::operator*() const [T = JSC::JSLock]
1   0x1418bd WTFCrash
2   0xb2c363 WTF::RefPtr<JSC::JSLock>::operator*() const
3   0xb2bf9c JSC::VM::apiLock()
4   0xb2ab0f JSC::JSLockHolder::init()
5   0xb2aabc JSC::JSLockHolder::JSLockHolder(JSC::ExecState*)
6   0xb2ab44 JSC::JSLockHolder::JSLockHolder(JSC::ExecState*)
7   0xed986 -[TNSArrayAdapter dealloc]
8   0x4802060 objc_object::sidetable_release(bool)
9   0x4801361 objc_release
10  0x480252d (anonymous namespace)::AutoreleasePoolPage::pop(void*)
11  0x6ce9c main
12  0x56aba21 start
13  0x1
```

src/NativeScript/ObjC/ObjCTypes.mm
```diff
@@ -202,35 +147,13 @@ JSValue toValue(ExecState* execState, id object, Structure* (^structureResolver)
...
-        if (JSObject* wrapper = valueWrapper.value) {
-            return wrapper;
-        }
-
-        ObjCWrapperObject* wrapper = ObjCWrapperObject::create(execState->vm(), structureResolver(), object, jsCast<GlobalObject*>(execState->lexicalGlobalObject()));
-        [TNSValueWrapper attachValue:wrapper
-                              toHost:object];
+    auto globalObject = jsCast<GlobalObject*>(execState->lexicalGlobalObject());
+    if (JSObject* wrapper = globalObject->interop()->objectMap().get(object)) {
         return wrapper;
```
jasssonpet: Consider enforcing the invariant that the `wrappedObject` is the same as `object` if possible.

src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
```diff
@@ -66,7 +67,7 @@ static void attachDerivedMachinery(GlobalObject* globalObject, Class newKlass, J
         ObjCSuperObject* superObject = ObjCSuperObject::create(vm, superStructure, derivedWrapper, globalObject);
         derivedWrapper->putDirect(vm, vm.propertyNames->superKeyword, superObject, ReadOnly | DontEnum | DontDelete);
 
-        [TNSValueWrapper attachValue:derivedWrapper toHost:instance];
+        globalObject->interop()->objectMap().set(instance, derivedWrapper);
```
jasssonpet: I wonder whether this `set` should have a matching `remove` call, like all the other wrappers.

fealebenpae: The destructor of `ObjCWrapperObject` removes itself from the object map.

src/NativeScript/ObjC/ObjCTypes.mm
fealebenpae: Should this be an `ASSERT` or `RELEASE_ASSERT`?

src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
jasssonpet: Oh, I've missed that. In that case would it be reasonable the constructor to add the object to the cache?

fealebenpae: I've thought about that - the designated initializer on the various adapters adds the object to the map and the `dealloc` method removes it from the map. I can do the same in `ObjCWrapperObject::finishCreation`.

jasssonpet:
:+1:

PR #452, opened 2015-12-23
Jasssonpet/require

jasssonpet:
Related to: https://github.com/NativeScript/common-runtime-tests-app/pull/11


fealebenpae:
Do we gain anything by replacing the `var`s with `let`s, given that we can't JIT?

jasssonpet:
Yes, it's easier to track the code flow without hoisting variables.

Issue #369, opened 2015-10-12
No corresponding source for the libffi build


mwcampbell:
There is no pointer to the source code for the specific version of libffi that is included in binary form in this repository, nor instructions to rebuild it. This means that if I find a problem with libffi, I can't debug it.

fealebenpae:
Hi @mwcampbell,

We haven't included the sources for libffi because we built it from two separate code bases. We built it once for arm64 with this revision https://github.com/atgreen/libffi/commit/536366349ced9d57c873d64e219ba30992a83c06 and we built it for armv7, i386 and x86_64 with this revision https://github.com/atgreen/libffi/commit/20562ac0427c3578250d04c6e34fb0127d4551cf (release 3.2.1). If we could automate the build of libffi we would have included it as a submodule like we did WebKit, but as it is right now we can't build armv7 and arm64 from the same revision.

The source isn't otherwise modified, so combining the builds of those two revisions in a fat binary will yield the same libffi library that's currently in the repo.

Unfortunately, debugging libffi right now means building the correct revision locally and replacing the production library in the repo with the debug build.

fealebenpae:
The issues that precluded libffi from building for armv7 and arm64 from the same revision have been resolved in our fork - https://github.com/nativescript/libffi.

PR #453, opened 2015-12-24
Libffi build

jasssonpet:
Remove the prebuilt binaries and add libffi build as dependency to the NativeScript target.

* As @fealebenpae suggested, this also allows iPhone 6 to run armv7 architecture (https://github.com/NativeScript/nativescript-cli/issues/1362)
* Allows stepping into libffi source

We should reset [libffi/master](https://github.com/NativeScript/libffi/tree/master) to [libffi/update](https://github.com/NativeScript/libffi/tree/update) before merging and tag it as well.

Related to: https://github.com/NativeScript/ios-runtime/issues/369


jasssonpet:
run ci

jasssonpet:
run ci

src/NativeScript/CMakeLists.txt
```diff
@@ -195,6 +195,12 @@ set_target_properties(NativeScript PROPERTIES
     "
 )
 
+add_custom_command(TARGET NativeScript
```
fealebenpae: I think it would be better to create a new custom target and add a dependency to it rather than use a custom build command on the runtime target.

jasssonpet: That was the first thing I did, but I couldn't find out how to pass the relative variables to the other target.

fealebenpae:
This looks great! I would like to point out this is excellent work, but I'm afraid @KristinaKoeva will have your hide for beating her to it :grin: 

As i pointed out inline, I think it's better to use a new target instead of a custom command.

CMakeLists.txt
```diff
@@ -12,7 +12,8 @@ set(CMAKE_XCODE_ATTRIBUTE_IPHONEOS_DEPLOYMENT_TARGET "7.0")
 
 set(CMAKE_MODULE_PATH "${CMAKE_SOURCE_DIR}/cmake")
 
-set(LIBFFI_DIR "${CMAKE_SOURCE_DIR}/src/libffi")
+set(LIBFFI_LIB_DIR "${CMAKE_SOURCE_DIR}/src/libffi/build_$(PLATFORM_NAME)-$(CURRENT_ARCH)/.libs")
```
fealebenpae: Maybe move these variables to `libffi.cmake`?

src/NativeScript/CMakeLists.txt
jasssonpet: Yeah, that is much better. I was missing the `SetActiveArchitectures` macro for that one to work :smile:

jasssonpet:
run ci

build/scripts/build-libffi.sh
```diff
@@ -0,0 +1,50 @@
...
+    else
+        echo "Invalid architecture: $ARCH"
+        exit 1
+    fi
+
+    autoreconf -i
+
+    mkdir -p "$BINARY_DIR" && pushd "$_"
+
+    local COMPILER_FLAGS="-arch $ARCH -isysroot $SDKROOT -miphoneos-version-min=$IPHONEOS_DEPLOYMENT_TARGET"
```
fealebenpae: I don't think it's a good idea to hard-code the deployment target flag, you should use Xcode's build settings to infer the flag, similar to https://github.com/NativeScript/ios-runtime/blob/master/build/scripts/buildWebKitFromWithinXcode.sh. That way we'll get automatic support for non-iPhoneOS platforms such as MacOSX or tvOS.

jasssonpet: Taking as much variables from Xcode environment is a great idea :smile: 

I've also edited the hardcoded optimization levels and debugging symbols flags.

jasssonpet:
@ns-bot retest this please

jasssonpet:
retest this please

jasssonpet:
run ci

Issue #450, opened 2015-12-23
Bad character in a comment will break require()


NathanaelA:
I have some code that ran on Android fine; pulled it down from the github repo to my Mac; attempted to run it and it kept crashing saying the variable wasn't defined.    Example:

```
var Me = require('./test.js');   
// makes Me === {}  if test.js contains a comment anywhere in it with the following: 
// ... don’t  ...
```
Some how the ' in don't is actually a high ascii character (0x92).  
This character causes the ios runtime to NOT return or execute the required file; and their is NO error!!!
Which of course lets your application to crash elsewhere in the process when it tries to do something with that required file...

Please note the required FILE actually doesn't run, I put a console.log in it at the very top of it, and it never outputted...

jasssonpet:
Hello, @NathanaelA,

I've tried reproducing this odd behavior with the following files:

```test.js
exports.x = 3;
// ... don’t  ...
exports.y = 4;
```

```index.js
var Me = require('./test.js');
console.dir(Me);
```

But it printed `x: 3, y: 4`. Maybe the special character was lost during copying. Can you confirm that we are talking about the `\u2019` character. Also if you could send me this as files, that would really help reproducing the issue.

Thank you!

NathanaelA:
Yeah, attached is a full test case. The only thing you need to do is install the ios runtimes.  It has everything else.  ;-)     I output everything to the console; and you will see the console in the test_bad.js doesn't run, nor does the module.exports get assigned.  
[testChar.zip](https://github.com/NativeScript/ios-runtime/files/70549/testChar.zip)

NathanaelA:
Oh and I realize what the character is; once I found it; it was simple to remove.  But it took a while to try and figure out why it wasn't working as this code had sat a while and I hadn't tested it since way before I did a major upgrade from Maveriks to El Capitan.     So I got to play process of elimination and I never expected the character to cause the iOS runtime to go wonky.

jasssonpet:
Indeed, saving the file with Windows-1252 encoding breaks the runtime.

I'll take mark it as a bug and we'll take a look at it.

fealebenpae:
[...]

PR #457, opened 2015-12-29
Throw an error if loading module source fails

fealebenpae:
Works around #450


jasssonpet:
Maybe we should do the same when reading `package.json` files?

fealebenpae:
@jasssonpet feel free to do so.

PR #458, opened 2015-12-30
Wrap bridged NSError objects in JavaScript Errors

fealebenpae:
Introduce the new `NSErrorWrapper` that will wrap an `NSError` thrown in JavaScript so that it can retain the `message` and `stack` properties expected in all JavaScript errors.

### DIFF
--- a/src/ts/Interop.d.ts
+++ b/src/ts/Interop.d.ts
@@ -206,4 +206,14 @@ declare module interop {
        */
       takeUnretainedValue(): T;
     }
+    
+    interface NSErrorWrapper extends Error {
+        error: NSError;
+    }
+    
+    declare var NSErrorWrapper: {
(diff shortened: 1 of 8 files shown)


src/ts/Interop.d.ts
```diff
@@ -206,4 +206,14 @@ declare module interop {
        */
       takeUnretainedValue(): T;
     }
+    
+    interface NSErrorWrapper extends Error {
+        error: NSError;
+    }
+    
+    declare var NSErrorWrapper: {
```
jasssonpet: Should this be `declare`-d? I'm receiving the following error:

> error TS1038: A 'declare' modifier cannot be used in an already ambient context.

fealebenpae: Thats's a good point - please feel free to fix it in-line.

PR #468, opened 2016-01-07
ES6 Modules

fealebenpae:
As the title suggests, this introduces support for the [ECMAScript 6 Modules syntax](http://www.2ality.com/2014/09/es6-modules-final.html). In fact, ES6 Modules are now first-class citizens, with CommonJS support implemented as a shim on top.


jasssonpet:
I think it's also safe to delete the BlankApp target.

fealebenpae:
@jasssonpet Er, why?

jasssonpet:
The whole purpose of the BlankApp was that the `require` routine isn't used. If I understand this correctly, now that everything is done in native, the first call from JavaScript files to native will always be from user code.

jasssonpet:
Also, we will have to update the `module.id` to be an absolute url, but we can do this after the merge to be more correct.

fealebenpae:
Well, yes, that's right, but BlankApp is still the target that just evaluates JavaScript without any module system. If that's not an important distinction then I'll remove it.

jasssonpet:
Ok, let's leave it for now.

src/NativeScript/Interop.mm
```diff
@@ -341,4 +338,12 @@ static EncodedJSValue JSC_HOST_CALL interopFuncBufferFromData(ExecState* execSta
 ErrorInstance* Interop::wrapError(ExecState* execState, NSError* error) const {
     return this->_nsErrorWrapperConstructor->createError(execState, error);
 }
+
+JSArrayBuffer* Interop::bufferFromData(ExecState* execState, NSData* data) const {
```
jasssonpet: Maybe this should be guarded too with `#ifdef`?

fealebenpae: You mean `#ifdef __OBJC__`?

examples/Gameraww/app/fetch.js
```diff
@@ -0,0 +1,15 @@
+module.exports = function fetch(path) {
```
jasssonpet: I think this would look better with ES6 modules syntax, like the other files.

fealebenpae: I explicitly left it like that so that Gameraww can exercise both the ES6 modules and the CommonJS shim.

jasssonpet: Well, we have plenty of tests for `require` calls, but as you wish.

fealebenpae: It's easier for me to develop and test out new things in Gameraww :grin: I suppose that as soon as modules are mature enough we can move this to ES6 as well.

src/NativeScript/GlobalObject.moduleLoader.mm
```diff
@@ -0,0 +1,388 @@
...
+    JSModuleRecord* moduleRecord = jsDynamicCast<JSModuleRecord*>(moduleRecordValue);
+    if (!moduleRecord) {
+        return jsUndefined();
+    }
+
+    GlobalObject* self = jsCast<GlobalObject*>(globalObject);
+    VM& vm = execState->vm();
+
+    if (JSValue moduleFunction = moduleRecord->getDirect(vm, self->_commonJSModuleFunctionIdentifier)) {
+        NSURL* moduleUrl = [NSURL fileURLWithPath:(NSString*)keyValue.toWTFString(execState).createCFString().get()];
```
jasssonpet: I wonder if this string should be autoreleased?

fealebenpae: `createCFString` returns a `WTF::RetainPtr<CFStringRef>` which releases the string as soon as it exits scope.

PR #471, opened 2016-01-11
Lock in TNSDictionaryAdapter keyEnumerator

jasssonpet:
Otherwise the "current thread is holding api lock" assert fails.


fealebenpae:
That's a good catch, but I think it's better to move the lock to `-[TNSDictionaryAdapter keyEnumerator]`, which calls the initializer.

jasssonpet:
@fealebenpae I've moved it there.

fealebenpae:
Looks good to me.

PR #472, opened 2016-01-11
Add events to timeline without implementing an inspector domain

KristinaKoeva:


fealebenpae:
Please delete `src/debugging/Inspector/scripts/generate.pyc` and add a `.gitignore` entry for `.pyc` files.

PR #475, opened 2016-01-12
Cache package.json resolution

jasssonpet:


fealebenpae:
Rename `requirePathCache` to something like `modulePathCache`, please. "require" is old news now :smiley:

src/NativeScript/GlobalObject.moduleLoader.mm
```diff
@@ -64,6 +64,11 @@
         absolutePath = [[absolutePath stringByAppendingPathComponent:path] stringByStandardizingPath];
     }
 
+    WTF::String requestedPath = absolutePath;
+    if (self->modulePathCache().contains(absolutePath)) {
```
fealebenpae: Use `requestedPath` instead of `absolutePath` here and on the next line - `requestPath` is already converted from an `NSString`.

PR #476, opened 2016-01-14
Swift-style initializers

fealebenpae:
This patch augments the JavaScript constructability of Objective-C classes. Constructor functions for Objective-C classes now support initialization via the Swift-style named parameter syntax which mimics Swift's initializer syntax.

Examples:
`NSURL.alloc().initWithString(aString)` becomes `new NSURL({ string: aString })`
`NSURL.alloc().initFileURLWithPath(aString)` becomes `new NSURL({ fileURLWithPath: aString })`
`NSString.alloc().initWithContentsOfFileEncodingError(aString, anEncoding)` becomes `new NSString({ contentsOfFile: aString, encoding: anEncoding })`

### DIFF
--- a/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
+++ b/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
@@ -12,6 +12,8 @@
 #include "ObjCWrapperObject.h"
 #include "ObjCMethodCall.h"
 #include "ObjCConstructorCall.h"
+#include "ObjCConstructorNative.h"
+#include "AllocatedPlaceholder.h"
 #include "ObjCTypes.h"
 #include "Interop.h"
 #include "PointerInstance.h"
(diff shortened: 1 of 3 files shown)


jasssonpet:
Is this change only about constructors or are methods going to follow the same pattern, too?

fealebenpae:
@jasssonpet only constructors, I'm not sure I can implement methods the same way.

jasssonpet:
run ci

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
```diff
@@ -146,8 +148,95 @@ static void writeAdapter(ExecState* execState, const JSValue& value, void* buffe
...
+        JSValue initializer;
+        if (JSFinalObject* argument = jsDynamicCast<JSFinalObject*>(execState->argument(0))) {
+            initializer = getInitializerForSwiftStyleConstruction(execState, constructor->instancesStructure()->storedPrototype(), computeInitializerJSName(execState, argument, initializerArguments));
+        } else if (execState->argumentCount() == 0) {
+            initializer = constructor->instancesStructure()->storedPrototype().get(execState, Identifier::fromString(execState, "init"));
+        } else {
+            goto new_legacy;
+        }
+
+        if (initializer && initializer.isCell()) {
```
ivanbuhov: If `getInitializerForSwiftStyleConstruction` doesn't find a suitable initializer corresponding to the given literal, do we want to fallback to the legacy `new` resolution or maybe throw an error (e.g. 'no initializer matching the given signature').
Is there any case in which `new` will be called with a literal (`FinalObject`) without targeting the new swift-style initializer resolution?

fealebenpae: You could conceivably invoke an initializer that accepts a single `NSDictionary` parameter and feed it an object literal using the legacy constructor resolution logic. I admit it's far-fetched, buuuuuut...

ivanbuhov: I agree, but hope the keys in the dictionary never match an initializer arguments names :)) ... yes, yes, this scenario is too far-fetched, and we have workaround ... but just saying :)

fealebenpae: I'm strongly in favor of removing the now-legacy constructor resolution in a couple of releases - it hardly ever works and is _slow_. Not to mention the enormous infrastructure required to support it.

ivanbuhov: :+1:

Issue #451, opened 2015-12-23
Allow inherited Objective-C classes in TypeScript to be minified


jasssonpet:
We currently register a native class with the name of the provided JavaScript class. However, during minification, class names become `a`, `b`, `c`... and sometimes collisions occur.

This is required for the Angular integration - https://github.com/NativeScript/nativescript-angular/wiki/Benchmarks#startup-time

fealebenpae:
I have an idea - let's deprecate inheriting from native classes in TypeScript's ES5 output and remove it in 2.0 :grin: 

But seriously, this causes problems every orher release.

jasssonpet:
We can also use the `--reserved` or `--reserved-file` flags of [UglifyJS](https://github.com/mishoo/UglifyJS2).

> ... you can use `--reserved-file` to pass a filename that should contain the names to be excluded from mangling. This file can be used both for excluding variable names and property names.

jasssonpet:
The implementation with custom static property name/decorator for the Objective-C class name seems to be not possible at the moment, because it isn't accessible by the time our custom `__extends` function is called.

PR #479, opened 2016-01-18
Rename Objective-C class name in case of extend conflicts

jasssonpet:
Close https://github.com/NativeScript/ios-runtime/issues/451.

### DIFF
--- a/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
+++ b/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
@@ -30,15 +30,14 @@ namespace NativeScript {
 using namespace JSC;
 using namespace Metadata;
 
-static WTF::CString generateAnonymousSubclassName(const char* className) {
-    int i = 0;
-    WTF::CString anonymousName;
+static WTF::CString computeRuntimeAvailableClassName(const char* userDesiredName) {
+    WTF::CString runtimeAvailableName(userDesiredName);
 
-    do {
-        anonymousName = WTF::String::format("%s%d", className, ++i).utf8();
-    } while (objc_getClass(anonymousName.data()));
+    for (int i = 1; objc_getClass(runtimeAvailableName.data()); ++i) {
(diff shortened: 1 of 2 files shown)


fealebenpae:
Looks good to me.

src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.mm
```diff
@@ -30,13 +30,12 @@ @protocol TNSDerivedClass
...
 
-static WTF::CString generateAnonymousSubclassName(const char* className) {
-    int i = 0;
-    WTF::CString anonymousName;
+static WTF::CString generateSafeSubclassName(const char* className) {
+    WTF::CString anonymousName(className);
 
-    do {
-        anonymousName = WTF::String::format("%s%d", className, ++i).utf8();
-    } while (objc_getClass(anonymousName.data()));
```
KristinaKoeva: I don't have a strong opinion on this one, but the previous ```while``` version looks more clear to me :)

jasssonpet: I don't mind returning it, but this way the `i` variable has more narrow scope.

KristinaKoeva: Do we all agree that this version is clear enough ? If so, leave it as is :)
ping @fealebenpae @ivanbuhov

fealebenpae: I believe its intent is clear enough, although I admit I am not unbiased in this. Maybe a small comment that explains what's a safe subclass name and why we need it could help, but I'm only offering this as a suggestion.

jasssonpet: @fealebenpae You are right, I've renamed the function to make its intent more clear.

fealebenpae: :+1:

Issue #95, opened 2015-04-28
Require for native APIs


PanayotCankov:
The global object is currently polluted for iOS with all classes, structs, constants, functions etc. etc. for all iOS SDKs and additional libraries included in an App.

We want to expose the native APIs through require. So for example the UIButton will not appear as a class available in the global object but rather should be pulled through `require("objc!UIKit")`.

We will try to keep that backward compatible and preserve the SDK APIs in the global object but we will try to promote the `require` usage as best practice.

moll:
Remember to perhaps scope or somehow differentiate the built-in module names of APIs from regular NPMs.

PR #485, opened 2016-01-21
Expose module specific global tables from metadata.

ivanbuhov:
First merge https://github.com/NativeScript/ios-metadata-generator/pull/42
This is needed in order to project clang modules as ES6 modules.


ivanbuhov:
This PR is a private spike related to https://github.com/NativeScript/ios-runtime/issues/95  which was not meant to be released.

PR #487, opened 2016-01-22
Pipe build logs through XCPretty

fealebenpae:


build/scripts/buildWebKitFromWithinXcode.sh
```diff
@@ -1,3 +1,3 @@
 #!/bin/bash
 
-xcodebuild -target JavaScriptCore -sdk "$SDKROOT" -configuration $CONFIGURATION ARCHS="$ARCHS" ONLY_ACTIVE_ARCH=$ONLY_ACTIVE_ARCH $DEPLOYMENT_TARGET_SETTING_NAME=${!DEPLOYMENT_TARGET_CLANG_ENV_NAME}
\ No newline at end of file
+xcodebuild -target JavaScriptCore -sdk "$SDKROOT" -configuration $CONFIGURATION ARCHS="$ARCHS" ONLY_ACTIVE_ARCH=$ONLY_ACTIVE_ARCH $DEPLOYMENT_TARGET_SETTING_NAME=${!DEPLOYMENT_TARGET_CLANG_ENV_NAME} GCC_WARN_INHIBIT_ALL_WARNINGS=YES
```
jasssonpet: @fealebenpae Does it make sense to add `xcpretty` here, too? This would reduce the build log in Xcode, that quite often makes it unresponsive.

fealebenpae: I'd rather not do that, because Xcode relies on error messages being printed in the format they are in so that they can be picked up by the issues navigator. Besides, with `GCC_WARN_INHIBIT_ALL_WARNINGS` set, the endless drove of warnings will cease and the output should be smaller. Also, you can switch away from the output tab while the build is underway to mitigate the unresponsiveness problem.

jasssonpet:
run ci

build/scripts/build.sh
```diff
@@ -68,8 +77,11 @@ echo "TNS_IPA_SIZE: "$GAMERAWW_IPA_SIZE"KB"
...
 echo "Building TestRunner..."
-xcodebuild -configuration Debug -sdk iphoneos -target TestRunner ARCHS="armv7" ONLY_ACTIVE_ARCH=NO  >> "$WORKSPACE/build.log" 2>&1
+xcodebuild_pretty -configuration Debug -sdk iphoneos -target TestRunner ARCHS="armv7" ONLY_ACTIVE_ARCH=NO
 echo "Packaging TestRunner..."
 xcrun -sdk iphoneos PackageApplication -v "$WORKSPACE/cmake-build/tests/TestRunner/Debug-iphoneos/TestRunner.app" \
     -o "$WORKSPACE/cmake-build/tests/TestRunner/Debug-iphoneos/TestRunner.ipa" \
-         >> "$WORKSPACE/build.log" 2>&1
\ No newline at end of file
+         >> "$WORKSPACE/build.log" 2>&1
+TestRunner/Debug-iphoneos/TestRunner.app" \
```
jasssonpet: What's the purpose of the following three lines? It seems to fail the build locally.

fealebenpae: Derp. Seems like faulty copy-pasta. Please get rid of it.

PR #495, opened 2016-02-01
Provide a way to execute inspector logic that should run before the first line of the application

KristinaKoeva:
In order to have source maps we needed a way to execute some java script that would return the source map content before the first application line. The end user should not see this so we hide it from the debugger. If you want to debug it you should pass a ТNS_dontSuppressAllPauses environment variable

### DIFF
--- a/src/NativeScript/TNSRuntime+Inspector.mm
+++ b/src/NativeScript/TNSRuntime+Inspector.mm
@@ -11,6 +11,7 @@
 #include <JavaScriptCore/APICast.h>
 #include <JavaScriptCore/JSONObject.h>
 #include <JavaScriptCore/Debugger.h>
+#include <JavaScriptCore/Completion.h>
 
 #include "GlobalObjectInspectorController.h"
 #import "TNSRuntime+Inspector.h"
@@ -49,14 +50,19 @@ private:
 
 - (instancetype)initWithRuntime:(TNSRuntime*)runtime
                  messageHandler:(TNSRuntimeInspectorMessageHandler)messageHandler;
+- (void)setup;
 
 @end
 
 @implementation TNSRuntime (Inspector)
 
 - (TNSRuntimeInspector*)attachInspectorWithHandler:(TNSRuntimeInspectorMessageHandler)messageHandler {
-    return [[[TNSRuntimeInspector alloc] initWithRuntime:self
-                                          messageHandler:messageHandler] autorelease];
+    TNSRuntimeInspector* runtimeInspector = [[TNSRuntimeInspector alloc] initWithRuntime:self
+                                                                          messageHandler:messageHandler];
+
+    [runtimeInspector setup];
+
+    return [runtimeInspector autorelease];
 }
 
 @end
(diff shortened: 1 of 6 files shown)


src/NativeScript/TNSRuntime+Inspector.mm
```diff
@@ -49,14 +50,19 @@ @interface TNSRuntimeInspector ()
...
 
 - (TNSRuntimeInspector*)attachInspectorWithHandler:(TNSRuntimeInspectorMessageHandler)messageHandler {
-    return [[[TNSRuntimeInspector alloc] initWithRuntime:self
-                                          messageHandler:messageHandler] autorelease];
+    TNSRuntimeInspector* runtimeInspector = [[[TNSRuntimeInspector alloc] initWithRuntime:self
+                                                                           messageHandler:messageHandler] autorelease];
+
+    [runtimeInspector setup];
+
+    return runtimeInspector;
```
fealebenpae: I think it's better to move the `autorelease` call here, because this is where the function returns.

src/NativeScript/inspector/SuppressAllPausesEnabler.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+SuppressAllPausesEnabler::~SuppressAllPausesEnabler() {
+    if (this->suppressAllPauses()) {
+        if (JSC::Debugger* debugger = this->m_globalObject.debugger()) {
+            debugger->setSuppressAllPauses(false);
+        }
+    }
+}
+
+bool SuppressAllPausesEnabler::suppressAllPauses() {
```
fealebenpae: This will work better as a static method. The environment variable is global so I don't think it makes sense to update a class field with its value.

src/NativeScript/inspector/SuppressAllPausesEnabler.cpp
```diff
@@ -0,0 +1,31 @@
...
+    static std::once_flag initializeEnvVariableOnceFlag;
+
+    std::call_once(initializeEnvVariableOnceFlag, [this]() {
+        this->m_dontSuppressAllPauses = getenv("NS_dontSuppressAllPauses");
```
fealebenpae: Let's make the prefix `TNS_` so that it matches the prefix of the runtime's public API.

src/NativeScript/TNSRuntime+Inspector.mm
```diff
@@ -89,6 +95,19 @@ - (instancetype)initWithRuntime:(TNSRuntime*)runtime
...
 
+- (void)setup {
+    WTF::Deque<WTF::RefPtr<JSC::Microtask>> other;
+    GlobalObject* globalObject = self->_runtime->_globalObject.get();
+
+    globalObject->microtasks().swap(other);
+
+    loadAndEvaluateModule(_runtime->_globalObject->globalExec(), WTF::ASCIILiteral("inspector_modules.js"));
+
+    JSC::JSLockHolder lock(_runtime->_vm.get());
```
fealebenpae: I think it's better to move this to the beginning of the method so that access to the microtasks collection is synchronized as well. Also, `loadAndEvaluateModule` might at some point expect that the API lock has been acquired.

fealebenpae:
Looks good to me.

Issue #259, opened 2015-08-07
More than one initializer found bug


NathanaelA:
Using v1.20 runtimes & v1.20 common, and v1.2.1 tns command.

Ok, this is a REALLY weird issue, and I've done as much debugging as I can at this point.   I've been working on a cool websocket library and now have it fully working on IOS and Android properly.  So I am trying to make it into a easy to use plugin and do some final tests; so I change
My applications require statement to
**require('nativescript-websockets');**
and the nativescript-websockets.js require to
**require('nativescript-websockets/websocket-common');**
and then install the plugin using the **tns plugin add nativescript-websockets**  

And everything installs properly and I see my directory and plugin in the platforms/ios/app/tns_modules folder like I expect.    

Now If I try to do **tns run ios --emulator** I get the following error message:
```
/app/tns_modules/file-system/file-system.js:314:24: JS ERROR Error: More than one initializer found that matches constructor invocation: initWithCapacity: initWithArray: initWithArray:copyItems: init
1   0xe7c30 NativeScript::FFICallback<NativeScript::ObjCMethodCallback>::ffiClosureCallback(ffi_cif*, void*, void**, void*)
2   0x6b1796 ffi_closure_SYSV_inner
3   0x6b11e2 .LCFI7
4   0x1710f24 -[UIViewController loadViewIfRequired]
5   0x1711215 -[UIViewController view]
6   0x1608215 -[UIWindow addRootViewControllerViewIfPossible]
7   0x1608689 -[UIWindow _setHidden:forced:]
8   0x1608940 -[UIWindow _orderFrontWithoutMakingKey]
9   0x1616c7d -[UIWindow makeKeyAndVisible]
10  0x6b10fd .LCFI1
11  0x6b16df ffi_call
12  0xe6511 NativeScript::ObjCMethodCall::derivedExecuteCall(JSC::ExecState*, unsigned char*)
13  0xe6808 long long NativeScript::FFICall::executeCall<NativeScript::ObjCMethodCall>(JSC::ExecState*)
14  0x45fc1e JSC::LLInt::setUpCall(JSC::ExecState*, JSC::Instruction*, JSC::CodeSpecializationKind, JSC::JSValue, JSC::LLIntCallLinkInfo*)
15  0x45c056 llint_slow_path_call
16  0x46497f llint_entry
17  0x4604c6 vmEntryToJavaScript
[...]

PR #507, opened 2016-02-11
Fix memory crashes

jasssonpet:
Probably related to: https://github.com/NativeScript/ios-runtime/issues/259


fealebenpae:
I don't think #259 is related, because it never crashed with junk collected memory.

But it doesn't hurt to have this, IMO, `JSC::DeferGCForAWhile` is cheap enough.

PR #512, opened 2016-02-12
Introduce pluggable module loaders

fealebenpae:
This patch adds the notion of module loaders, similar to [RequireJS Plugins](http://requirejs.org/docs/plugins.html).

Module loaders introduce new syntax in module identifiers: `import x from '@foo/bar'` will try to import a module called `foo-module-loader` and defer the `resolve`, `fetch`, `translate`, `instantiate` and `evaluate` steps to it by invoking each exported hook function and passing the original `@foo/bar` key.

### DIFF
--- a/examples/Gameraww/app/tns_modules/uiimage-module-loader/index.js
+++ b/examples/Gameraww/app/tns_modules/uiimage-module-loader/index.js
@@ -1,0 +1,21 @@
+export function resolve(loader, key, referer) { return key; }
+
+export function fetch(loader, key) {
+  let imageName = key.replace(/^@.*\//, '');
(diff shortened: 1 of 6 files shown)


examples/Gameraww/app/tns_modules/uiimage-module-loader/index.js
```diff
@@ -0,0 +1,21 @@
+export function resolve(loader, key, referer) { return key; }
+
+export function fetch(loader, key) {
+  let imageName = key.replace(/^@.*\//, '');
```
jasssonpet: I think that the regex should be `^@.*?\/` to make sure `@uiimage/folder/reddit-default` is matched correctly.

PR #513, opened 2016-02-16
Name conflict tests

KristinaKoeva:
When we see a method from a base class with the same name as a property in a derived class, we change the method metadata in the base class to property metadata, so tests that assert this behavior are added.

### DIFF
--- a/tests/TestRunner/app/Inheritance/InheritanceTests.js
+++ b/tests/TestRunner/app/Inheritance/InheritanceTests.js
@@ -1846,7 +1846,7 @@ describe(module.id, function () {
         expect(NSString.stringWithCString(encoding).toString()).toBe("v@:");
     });
 
-    it("should project Symbol.iterable as NSFastEnumeration", function () {
+    it("should project Symbol.iterable as NSFastEnumeration", function() {
         var start = 1;
         var end = 10;
         var lastStep = 0;
@@ -1855,15 +1855,16 @@ describe(module.id, function () {
                 return {
                     step: start,
 
-                    next() {
+                        next() {
                         if (this.step <= end) {
-                            return { value: this.step++, done: false };
+                            return { value : this.step++, done : false };
                         } else {
-                            return { done: true };
+                            return { done : true };
                         }
-                    },
+                    }
+                    ,
 
-                    return() {
+                        return () {
                         lastStep = this.step;
                         return {};
                     }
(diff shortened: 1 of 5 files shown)


tests/TestRunner/app/Inheritance/InheritanceTests.js
```diff
@@ -1846,39 +1846,54 @@ describe(module.id, function () {
...
-        TNSClearOutput();
-    })
+      expect(IterableClass.conformsToProtocol(NSFastEnumeration)).toBe(true);
+      expect(actual).toEqual(expected);
+      expect(lastStep).toEqual(end + 1);
+
+      TNSClearOutput();
+    });
+
+    it("Mehtod and property with the same name", function() {
```
jasssonpet: `Mehtod` -> `Method` :smile:

tests/TestRunner/app/Inheritance/InheritanceTests.js
```diff
@@ -1846,39 +1846,54 @@ describe(module.id, function () {
...
+      var JSObject =
+          NSObject.extend({get conflict() { return true; }},
+                          {protocols : [ TNSPropertyMethodConflictProtocol ]});
+
+      var derived = JSObject.new();
+      var result;
+      expect(function() {
+        result = TNSTestNativeCallbacks.protocolWithNameConflict(derived);
+      }).not.toThrow();
```
jasssonpet: I think that this expectation is redundant - all calls shouldn't throw an error and the test will fail if such happens.

tests/TestRunner/app/Inheritance/InheritanceTests.js
```diff
@@ -1846,39 +1846,54 @@ describe(module.id, function () {
...
-                            return { done: true };
-                        }
-                    },
-
-                    return() {
-                        lastStep = this.step;
-                        return {};
-                    }
-                }
+    it("should project Symbol.iterable as NSFastEnumeration", function() {
```
jasssonpet: Somehow the indentation got mixed here and git is marking it as changed.

KristinaKoeva: The clang format on my machine formatted it that way. :rose: 
I will execute it again to see if this is the right formatting

jasssonpet: I think we don't use clang-format for JS files: https://github.com/NativeScript/ios-runtime/blob/e34fe1094df0868eb34147c06276b421fffb0493/tools/pre-commit-clang-format#L33

But if this is the right indentation, I'm all in for that.

KristinaKoeva: I am not using the pre commit hook, but execute clang format manually, so that's why they were changed :)

PR #514, opened 2016-02-16
Unify initializers generation logic in ObjCConstructorBase

ivanbuhov:
1. Unify initializers generation logic in `ObjCConstructorNative` and `ObjCConstructorDerived`. Conceptually, it is the same, so there is no reason to be in two places.
2. `ObjCConstructor`s don't need `InterfaceMeta` to be initialized. They already require a prototype from where they can get the `InterfaceMeta`.
3. `ObjCConstructorDerived` don't need a parent constructor to be initialized, anymore.

Also, this makes `ObjCConstructorDerived` class almost identical to its base class (`ObjCConstructorBase`). This will allow us to remove it in the future.

### DIFF
--- a/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.h
+++ b/src/NativeScript/ObjC/Constructor/ObjCConstructorBase.h
@@ -12,6 +12,10 @@
 #include <functional>
 #include "FFIType.h"
 
+namespace Metadata {
+struct InterfaceMeta;
+}
+
 namespace NativeScript {
 class ObjCConstructorCall;
 
@@ -19,15 +23,12 @@ class ObjCConstructorBase : public JSC::InternalFunction {
 public:
     typedef JSC::InternalFunction Base;
 
+    static const unsigned StructureFlags;
+
     DECLARE_INFO;
 
     Class klass() const {
         return this->_klass;
-    }
-
-    const WTF::Vector<ObjCConstructorCall*> generateInitializers(JSC::VM& vm, GlobalObject* globalObject, Class target) {
-        JSC::DeferGCForAWhile deferGC(vm.heap);
-        return this->_initializersGenerator(vm, globalObject, target);
     }
 
     JSC::Structure* instancesStructure() const {
@@ -41,6 +42,8 @@ public:
     static WTF::String className(const JSObject* object);
 
     const WTF::Vector<JSC::WriteBarrier<ObjCConstructorCall>>& initializers(JSC::VM&, GlobalObject*);
+
+    const Metadata::InterfaceMeta* metadata();
 
 protected:
     ObjCConstructorBase(JSC::VM& vm, JSC::Structure* structure)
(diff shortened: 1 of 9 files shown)


src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
```diff
@@ -132,9 +150,23 @@ static void writeAdapter(ExecState* execState, const JSValue& value, void* buffe
...
     if (this->_initializers.size() == 0) {
-        for (ObjCConstructorCall* initializer : this->generateInitializers(vm, globalObject, this->klass())) {
-            this->_initializers.append(WriteBarrier<ObjCConstructorCall>(vm, this, initializer));
-        }
+        JSC::DeferGCForAWhile deferGC(vm.heap);
+        const Metadata::InterfaceMeta* metadata = this->metadata();
+
+        do {
+            std::vector<const Metadata::MethodMeta*> initializers = metadata->initializersWithProtcols();
+            for (std::vector<const Metadata::MethodMeta*>::iterator init = initializers.begin(); init != initializers.end(); ++init) {
```
jasssonpet: You can use `for (const Metadata::MethodMeta* method : initializers)` if you want.

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
```diff
@@ -13,10 +13,12 @@
 #include "ObjCMethodCall.h"
 #include "ObjCConstructorCall.h"
 #include "ObjCConstructorNative.h"
+#include "../ObjCPrototype.h"
```
jasssonpet: You can remove the here (and bellow) `../` to be consistent with other header includes. Headers are searched recursively in all source directories.

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.h
```diff
@@ -42,9 +41,12 @@ class ObjCConstructorBase : public JSC::InternalFunction {
 
     const WTF::Vector<JSC::WriteBarrier<ObjCConstructorCall>>& initializers(JSC::VM&, GlobalObject*);
 
+    const Metadata::InterfaceMeta* metadata();
+
 protected:
     ObjCConstructorBase(JSC::VM& vm, JSC::Structure* structure)
-        : Base(vm, structure) {
+        : Base(vm, structure)
+        , _metadata(nullptr) {
```
jasssonpet: You can move the field initialization to the other ones in `finishCreation`.

jasssonpet:
I think you can also remove the structure flags of the derived and native constructors and move the  `JSC::OverridesGetOwnPropertySlot` flag to the base class, since that's where the `getOwnPropertySlot` override is.

src/NativeScript/ObjC/Constructor/ObjCConstructorBase.mm
```diff
@@ -88,6 +90,22 @@ static void writeAdapter(ExecState* execState, const JSValue& value, void* buffe
     return "@";
 }
 
+const Metadata::InterfaceMeta* ObjCConstructorBase::metadata() {
+    if (this->_metadata == nullptr) {
+        JSObject* proto = this->_prototype.get();
+        ObjCPrototype* objcPrototype = nullptr;
+        while (proto && !(objcPrototype = jsDynamicCast<ObjCPrototype*>(proto))) {
+            proto = (proto->prototype().isCell()) ? jsDynamicCast<ObjCPrototype*>(proto->prototype().asCell()) : nullptr;
```
jasssonpet: I think this is best written as `proto = jsDynamicCast<JSObject*>(proto->prototype());` since `jsDynamicCast` is already checking for cells and stuff.

PR #515, opened 2016-02-17
Make tests run on iOS 8

jasssonpet:


tests/TestRunner/app/ObjCConstructors.js
```diff
@@ -79,46 +79,52 @@ describe("Constructing Objective-C classes with new operator", function () {
...
+                otherButtonTitles: null
+            });
+            expect(alertView.title).toEqual("About");
+            expect(alertView.message).toEqual("NativeScript Team");
+            expect(alertView.buttonTitleAtIndex(0)).toEqual("OK");
+        });
+
+        it("should support void initializers", () => {
+            let transform = new TNSCInterface({
+                identity: void 0
```
fealebenpae: Shouldn't this be `empty: void 0`, seeing as how the initializer is named `initWithEmpty`?

jasssonpet: This sure is so, but I wonder how the `initWithStructure:0.0 called` gets called. We should investigate this further :dizzy_face:

PR #517, opened 2016-02-19
Throw error when overriding properties without getter/setter.

jasssonpet:
Throw error when overriding a (non-readonly) native property without both a getter/setter function:

```objective-c
@interface TNSBaseInterface : NSObject
@property int baseProperty;
@end
```
```javascript
TNSBaseInterface.extend({
    get baseProperty() {
        console.log('get');
        return this.super.baseProperty;
    },
    // set baseProperty(value) {
    //     console.log('set');
    //     this.super.baseProperty = value;
    // },
});
```


fealebenpae:
:+1: It is right to protect this condition with an error so that the runtime doesn't crash.

PR #525, opened 2016-03-07
Provide runtime implementation for source maps

KristinaKoeva:
This is necessary for the scenarios where somebody wants to use the runtime without the modules. This agent will be overridden when the modules provide custom network domain implementation


ivanbuhov:
:+1:

PR #540, opened 2016-03-23
Throw error on unknown module character encoding

jasssonpet:
Closes https://github.com/NativeScript/ios-runtime/issues/539


ivanbuhov:
:+1:

PR #542, opened 2016-04-06
Retrieve inspector version from package.json

### LINKED ISSUE #529: Release Inspector with same version as {N} release
Now version (what is listed in about dialog) of inspector is:
Version 1.0 (1)

May be it sill be better if we use version same as {N} release.

KristinaKoeva:
Fixes #529

### DIFF
--- a/build/scripts/build-inspector.sh
+++ b/build/scripts/build-inspector.sh
@@ -8,6 +8,7 @@ MACOSX_DEPLOYMENT_TARGET="10.10"
 
 WEBKIT_SOURCE_PATH="$WORKSPACE/src/webkit"
 WEBKIT_BUILD_OUTPUT_PATH="$WORKSPACE/cmake-build/WebKit-Xcode"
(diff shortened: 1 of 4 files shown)


build/scripts/build-inspector.sh
```diff
@@ -41,7 +42,10 @@ find "$WEBKIT_BUILD_OUTPUT_PATH/Build/Products/$CONFIGURATION" -name "*.framewor
 
 checkpoint "Building Inspector app..."
 rm -rf "$INSPECTOR_BUILD_OUTPUT_PATH"
+
+version=$(python "$BUILD_DIR/scripts/get_version.py" "$BUILD_DIR/npm/inspector_package.json" 2>&1)
```
jasssonpet: Maybe `node -e "console.log(require('./inspector_package').version)"` could be easier.

KristinaKoeva: I decided to reuse the python logic, but it's OK for me to change it if you wish :)

PR #561, opened 2016-05-16
Fix debugging for code that was already executed

### LINKED ISSUE #315: Debugger does not stop at some breakpoints when --start option is in use on iOS7
Tested on iPad iOS7, not reproducible on iOS8 device.

Scenario - I use Cuteness app for testing purposes.
I execute `tns debug ios` which launches the debugger with the default *--debug-brk* option. Than I put two breakpoints in main-page.js file:
- 1st breakpoint at pageLoaded() function,
- 2nd breakpoint at listViewItemTap() function.

To verify that the debugger hits the breakpoints I tap on any item in the list (2nd breakpoint) and navigate back (1nd breakpoint) - it works as expected.

The issue is when I execute `tns debug ios --start` in order to attach to the running application. When open an item and go back the debugger stops at the 1st breakpoint, but it does not at the 2nd.

KristinaKoeva:
Typical iOS application is always in an entry scope, because the UIApplicationMain method never finishes. On debugger attach the debugger agent tries to recreate the bytecode so that it could insert all opt debug codes. Since the vm is in an entry scope the recreation never happens. So to workaround this problem we execute all listeners when the run loop is idle and if the current entry scope is the UIApplicationMain

Fixes #315

### DIFF
--- a/src/NativeScript/GlobalObject.mm
+++ b/src/NativeScript/GlobalObject.mm
@@ -7,45 +7,46 @@
 //
 
 #include "GlobalObject.h"
-#include <string>
-#include <JavaScriptCore/FunctionPrototype.h>
-#include <JavaScriptCore/FunctionConstructor.h>
-#include <JavaScriptCore/Microtask.h>
-#include <JavaScriptCore/Completion.h>
-#include <JavaScriptCore/StrongInlines.h>
-#include <JavaScriptCore/JSGlobalObjectFunctions.h>
-#include <JavaScriptCore/runtime/JSConsole.h>
-#include <JavaScriptCore/inspector/JSGlobalObjectConsoleClient.h>
-#include "ObjCProtocolWrapper.h"
-#include "ObjCConstructorNative.h"
-#include "ObjCPrototype.h"
-#include "Metadata.h"
-#include "SymbolLoader.h"
+#include "AllocatedPlaceholder.h"
+#include "FFICallPrototype.h"
 #include "FFIFunctionCall.h"
-#include "RecordConstructor.h"
-#include "RecordPrototypeFunctions.h"
+#include "FFIFunctionCallback.h"
 #include "Interop.h"
-#include "inspector/GlobalObjectInspectorController.h"
-#include "ObjCExtend.h"
-#include "ObjCTypeScriptExtend.h"
-#include "__extends.h"
-#include "ObjCMethodCall.h"
-#include "ObjCConstructorCall.h"
-#include "ObjCConstructorDerived.h"
+#include "JSWeakRefConstructor.h"
+#include "JSWeakRefInstance.h"
+#include "JSWeakRefPrototype.h"
+#include "Metadata.h"
 #include "ObjCBlockCall.h"
 #include "ObjCBlockCallback.h"
-#include "ObjCMethodCallback.h"
-#include "FFIFunctionCallback.h"
-#include "JSWeakRefConstructor.h"
-#include "JSWeakRefPrototype.h"
-#include "JSWeakRefInstance.h"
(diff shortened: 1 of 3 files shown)


src/NativeScript/GlobalObject.mm
```diff
@@ -7,45 +7,46 @@
 //
 
 #include "GlobalObject.h"
-#include <string>
```
KristinaKoeva: clang format on my machine rearranged those but if you'd like I will bring them back as they used to be 🌹

jasssonpet: This looks good to me 😄 

Maybe we should reformat all sources like this, but this is for another time.

src/NativeScript/GlobalObject.mm
```diff
@@ -118,6 +119,27 @@ static void microtaskRunLoopSourcePerformWork(void* context) {
     self->drainMicrotasks();
 }
 
+static void runLoopBeforeWaitingPerformWork(CFRunLoopObserverRef observer, CFRunLoopActivity activity, void* info) {
+    GlobalObject* self = static_cast<GlobalObject*>(info);
+    VMEntryScope* currentEntryScope = self->vm().entryScope;
+    if (self->vm().topCallFrame && currentEntryScope && !currentEntryScope->didPopListeners().isEmpty()) {
+        FFIFunctionCall* function_call = static_cast<FFIFunctionCall*>(self->vm().topCallFrame->callee());
+        const Meta* meta = Metadata::MetaFile::instance()->globalTable()->findMeta(WTF::ASCIILiteral("UIApplicationMain"));
```
ivanbuhov: 👍 
PS: `findMeta` can be called with pure `char *`, no need to construct `ASCIILiteral`.

src/NativeScript/GlobalObject.mm
```diff
@@ -118,6 +119,27 @@ static void microtaskRunLoopSourcePerformWork(void* context) {
     self->drainMicrotasks();
 }
 
+static void runLoopBeforeWaitingPerformWork(CFRunLoopObserverRef observer, CFRunLoopActivity activity, void* info) {
+    GlobalObject* self = static_cast<GlobalObject*>(info);
+    VMEntryScope* currentEntryScope = self->vm().entryScope;
+    if (self->vm().topCallFrame && currentEntryScope && !currentEntryScope->didPopListeners().isEmpty()) {
+        FFIFunctionCall* function_call = static_cast<FFIFunctionCall*>(self->vm().topCallFrame->callee());
```
jasssonpet: I think that you can use `jsDynamicCast()` here. That will make sure that the function call is indeed a `FFIFunctionCall` or return null otherwise so that it can be checked below.

src/NativeScript/GlobalObject.mm
```diff
@@ -118,6 +119,27 @@ static void microtaskRunLoopSourcePerformWork(void* context) {
     self->drainMicrotasks();
 }
 
+static void runLoopBeforeWaitingPerformWork(CFRunLoopObserverRef observer, CFRunLoopActivity activity, void* info) {
```
jasssonpet: Is this function supposed to be executed before the processing of the run loop? If this is the case, is it worth it to check that the activity is an entry one.

jasssonpet:
👍

src/NativeScript/GlobalObject.mm
KristinaKoeva: The activity is specified when we create the observer so this should be called only on kCFRunLoopBeforeWaiting
https://github.com/NativeScript/ios-runtime/pull/561/files#diff-c94a128c20882442378c1dfa04930835R213

PR #573, opened 2016-05-31
Attaching a debugger crashes the app if it was already livesynced

### LINKED ISSUE #572: Attaching a debugger crashes the app if it was already livesynced
Here are the steps that crash the app:
* Creating a simple helloworld app: `tns create nsHelloWorld`
* Deploy to the device using: `tns deploy ios`
* Livesync the app on to the device using: `tns livesync ios`
* Attach the debugger to the running app: `tns debug ios --start`

The app crashes with the following stacktrace and the debugger doesn't work:
image

This is a major issue in the **AppBuilder** clients as the livesync feature is enabled by default and is happening seamlessly in the background. Thus the {N} app in AppBuilder will crash most of the time when debugger is attached.

KristinaKoeva:
The app is initialized with the special live sync path that contains a space so URLWithString returns nil.

Fixes #572


KristinaKoeva:
run ci

KristinaKoeva:
run ci

PR #585, opened 2016-06-22
Change IPHONEOS_DEPLOYMENT_TARGET in project template to 8.0

ivanbuhov:


jasssonpet:
My Xcode (7.3) is adding `IPHONEOS_DEPLOYMENT_TARGET = 8.0;` to line 348 to match the version on line 329. I wonder if we should add it there, too.

PR #586, opened 2016-06-23
Report to the user that LiveSync contents is being unzipped

jasssonpet:
The application is displaying a black screen for 15 seconds at worst.

ping @tzraikov


tzraikov:
👍

PR #587, opened 2016-06-23
Remove the static library from npm package

ivanbuhov:
Remove the static library from npm package but still use it for example apps.


jasssonpet:
I guess we can also remove the explicit building of the static library here: https://github.com/NativeScript/ios-runtime/blob/edce27cf178cf060d9f25079e32ad64f640dc3e1/build/scripts/build.sh#L39-L53

It will still get built for the example apps, but we'll skip the iPhone Simulator architectures.

build/scripts/build-runtime.sh
```diff
@@ -17,10 +17,8 @@ cp -R "$DIST_DIR/TKLiveSync.framework" "$INTERNAL_DIR"
 
 . "$WORKSPACE/build/scripts/build.sh"
 
-mkdir -p "$INTERNAL_DIR/NativeScript/Frameworks"
-
-cp -R "$DIST_DIR/NativeScript/" "$INTERNAL_DIR/NativeScript"
-cp -R "$DIST_DIR/NativeScript.framework" "$INTERNAL_DIR/NativeScript/Frameworks"
+mkdir -p "$INTERNAL_DIR/NativeScript"
+cp -R "$DIST_DIR/NativeScript.framework" "$INTERNAL_DIR/NativeScript/NativeScript.framework"
```
jasssonpet: Let's drop the `NativeScript` framework now that it contains only the framework, what do you think?

PR #592, opened 2016-06-30
Changed the livesync port to enable simultaneous ios/android livesync

tzraikov:


jasssonpet:
run ci

jasssonpet:
Please note that we don't have a Inspector PR build. Maybe @dtopuzov may help with that.

KristinaKoeva:
Please correct me if I am wrong but as far as I understand the Android runtime is exposing a named socket and later CLI forwards it to a local tcp socket on port 18181. Take a look here - https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/livesync/android-livesync-service.ts#L56

So instead of changing the iOS runtime port, is it possible to just change the value of this one https://github.com/NativeScript/nativescript-cli/blob/master/lib/services/livesync/android-livesync-service.ts#L10 ?

PR #608, opened 2016-07-27
Use libzip library in TKLiveSync plugin

jasssonpet:
It is 10% faster when unzipping and provides more granular access to zip contents.

ping @tzraikov


tzraikov:
👍

Issue #626, opened 2016-08-16
Inspector crashes when Debug + LiveSync command executes on iOS device


vchimev:
# How to reproduce

A connected device is required, for example:
```
mcsofvchimev:appJS vchimev$ tns device
┌───┬─────────────────────────┬──────────┬──────────────────────────────────────────┬────────┬───────────┐
│ # │ Device Name             │ Platform │ Device Identifier                        │ Type   │ Status    │
│ 1 │ iPhone 5s iOS 93 Device │ iOS      │ ccd596fa85d0c0c78c9b5a95147c1ba49c9dcd67 │ Device │ Connected │
└───┴─────────────────────────┴──────────┴──────────────────────────────────────────┴────────┴───────────┘
```

Command to reproduce:
```
tns debug ios --watch
```

# Result

NativeScript Inspector fails to load the resources.

Output: 
```
mcsofvchimev:appJS vchimev$ tns debug ios --watch
Project successfully prepared (ios)
Transferring project files...
Successfully transferred all files.
Applying changes...
Mounting /Applications/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/DeviceSupport/9.3 (13E230)/DeveloperDiskImage.dmg
[Mounting] LookingUpImage
Successfully run application org.nativescript.appJS on device with ID ccd596fa85d0c0c78c9b5a95147c1ba49c9dcd67.
2016-08-16 11:43:52.781 appJS[1251:644979] Unzipping LiveSync folder. This could take a while...
2016-08-16 11:43:52.802 appJS[1251:644979] Unzipped 8 entries in 15.855014ms.
2016-08-16 11:43:52.803 appJS[1251:644979] tns_modules folder not livesynced. Using tns_modules from the already deployed bundle...
NativeScript loaded bundle file:///System/Library/Frameworks/UIKit.framework
NativeScript loaded bundle file:///System/Library/Frameworks/Foundation.framework

Setting up proxy...
Press Ctrl + C to terminate, or disconnect.

Successfully synced application org.nativescript.appJS on device ccd596fa85d0c0c78c9b5a95147c1ba49c9dcd67.
NativeScript loaded bundle file:///System/Library/Frameworks/CoreImage.framework
NativeScript loaded bundle file:///System/Library/Frameworks/QuartzCore.framework
NativeScript loaded bundle file:///System/Library/Frameworks/iAd.framework
NativeScript loaded bundle file:///System/Library/Frameworks/AVFoundation.framework
[...]

PR #634, opened 2016-08-30
Handle the case when we have tns_modules as symbolic link

KristinaKoeva:
The NSDirectoryEnumerator does not enumerate symbolic links, so when we have a symbolic link call the same method with the original file location

#626

### DIFF
--- a/src/NativeScript/inspector/CachedResource.mm
+++ b/src/NativeScript/inspector/CachedResource.mm
@@ -2,26 +2,50 @@
 #include "MimeTypeHelper.h"
 
 namespace Inspector {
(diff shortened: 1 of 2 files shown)


src/NativeScript/inspector/CachedResource.mm
```diff
@@ -2,26 +2,50 @@
 #include "MimeTypeHelper.h"
 
 namespace Inspector {
+Inspector::CachedResource createCacheResource(WTF::HashMap<WTF::String, Inspector::CachedResource>& cachedResources, NSString* basePath, NSString* filePath, NSString* prefix) {
```
jasssonpet: Can these functions be made static? Also, the `cachedResources` argument seems unused.

PR #645, opened 2016-09-08
Don't try to live edit if there is no module with such id. Also combine the absolute path because CLI doesn't know it

KristinaKoeva:


src/NativeScript/inspector/GlobalObjectDebuggerAgent.mm
```diff
@@ -72,12 +72,20 @@ void GlobalObjectDebuggerAgent::enable(ErrorString& errorString) {
...
+
     JSValue registry = this->m_globalObject->moduleLoader()->get(this->m_globalObject->globalExec(), Identifier::fromString(&this->m_globalObject->vm(), "registry"));
     JSMap* map = jsCast<JSMap*>(registry);
-    JSValue value = map->get(this->m_globalObject->globalExec(), JSC::jsString(&this->m_globalObject->vm(), scriptIdStr));
+    WTF::String scriptAbsolutePath = [NSString pathWithComponents:@[ this->m_globalObject->applicationPath(), @"app", scriptIdStr ]];
+    JSValue value = map->get(this->m_globalObject->globalExec(), JSC::jsString(&this->m_globalObject->vm(), scriptAbsolutePath));
     Identifier moduleIdentifier = Identifier::fromString(&this->m_globalObject->vm(), "module");
 
+    if (value.isUndefined()) {
+        error = ASCIILiteral("Could not find module at path: '%s'") + scriptAbsolutePath;
```
jasssonpet: Shouldn't this be either string format or string append?

KristinaKoeva: No, I decided to take the best of the two :D
Will fix it right away :)

PR #660, opened 2016-09-27
Makes the metadata generation build step compatible with Python 3

ivanbuhov:
Handle [change in the behavior of `map()` function](https://docs.python.org/3.0/whatsnew/3.0.html#views-and-iterators-instead-of-lists) in Python 3.

### DIFF
--- a/build/scripts/build-step-metadata-generator.py
+++ b/build/scripts/build-step-metadata-generator.py
@@ -6,6 +6,8 @@ import shlex
 import sys
 
 
+print("Python version: " + sys.version)
+
 def env(env_name):
     return os.environ[env_name]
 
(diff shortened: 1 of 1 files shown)


jasssonpet:
What was the behavior before this change? At NativeScript developer day there was a person that had Python 3 and the metadata generation for i386 went fine, but for x86_64 it seemed to strip the preprocessor definitions through the map function. I wonder if this is the same here.

ivanbuhov:
@jasssonpet Correct. This was the buggy behavior. You can find more info here: https://github.com/NativeScript/NativeScript/issues/2625. `map()` in Python 3 return iterators instead of list.

jasssonpet:
👍

Issue #607, opened 2016-07-26
Update the sample with the embedded iOS runtime


ivanbuhov:
Currently, [this sample](https://github.com/NativeScript/sample-ios-embedded) is crashing with iOS runtime v2.1.0 and later (consider setupping an automated build).

ivanbuhov:
Fixed with https://github.com/NativeScript/sample-ios-embedded/pull/1 and https://github.com/NativeScript/ios-runtime/pull/661

PR #661, opened 2016-09-27
Export all public symbols from JavaScriptCore C API

ivanbuhov:
Requirement for https://github.com/NativeScript/ios-runtime/issues/607
Also related to: https://github.com/NativeScript/sample-ios-embedded/pull/1


jasssonpet:
Can we somehow export all symbols marked with `JS_EXPORT` instead?

ivanbuhov:
@jasssonpet According to [this guy](http://stackoverflow.com/a/32548642/3968054) there is no easy way for a dynamic library to automatically reexport symbols defined in a static library after linking against it.
A possible solution is to filter only needed symbols with the help of `Clang`/`metadata generator` or using `nm` and some regex magic and dump the result in a [exports file](https://developer.apple.com/library/content/documentation/MacOSX/Conceptual/BPFrameworks/Tasks/ExportingInterfaces.html).
Since this is not related to https://github.com/NativeScript/ios-runtime/issues/607 I suggest to treat it as a separate issue.

fealebenpae:
You could, I don't know, use this [list of symbols JavaScriptCore exports](https://github.com/WebKit/webkit/blob/master/Source/JavaScriptCore/JavaScriptCore.order) 😉

jasssonpet:
We can export the full public JSC API: https://gist.github.com/anonymous/aab759df6c372085c72c214acc95b575

What do you think about building the JavaScriptCore as a shared framework then.

fealebenpae:
Building as a framework would needlessly complicate matters with private headers and fat binaries. IMO you can even manually hardcode the list of symbols in the public C API - they are few enough and a new one gets added once in a blue moon. But personally I would just use the order file.

ivanbuhov:
@fealebenpae It seems that not all of the symbols in the list are included in the static binary. For example `JSC::ArrayProfile::computeUpdatedPrediction(JSC::CodeBlock*, JSC::OperationInProgress)` is present in the `JavaScriptCore.order` file but there is no such method in the source and no such symbol available in the static library. I guess there is a good reason for these inconsistencies and maybe we can handle them in such a way but still IMO exporting all APIs included in the list is a bit overkill.

Since we have no use cases that require dynamic library nor exporting all of the JSC APIs I am voting for manually listing only the public JSC APIs :)

jasssonpet:
You can put them in a separate file if you'd like similarly to https://github.com/NativeScript/ios-runtime/blob/buhov/embeding-ios-runtime-fix/tests/TestFixtures/exported-symbols.txt

jasssonpet:
run ci

jasssonpet:
run ci

jasssonpet:
run ci

pkoleva:
run ci

pkoleva:
run ci

pkoleva:
run ci

jasssonpet:
run ci

PR #674, opened 2016-10-21
Build test app in release config

jasssonpet:
Reuse the release config runtime cache to speed up the builds.

This will remove all asserts that are enabled in debug config only from the runtime and it's dependencies (metadata generator, WebKit, libffi).


jasssonpet:
run ci

jasssonpet:
run ci

PR #671, opened 2016-10-17
Update webkit

KristinaKoeva:

PR #687, opened 2016-11-09
Prevent crashing in StackVisitor::readFrame if built in release

ivanbuhov:
After [updating webkit](https://github.com/NativeScript/ios-runtime/pull/671) the runtime started to crash randomly in `StackVisitor::readFrame` in case it is built for release. It seems that the top frame of the call stack is the buggy one which causes the crash. Setting `useCurrentFrame` to `false` on error creation skips the top call frame(the buggy one) when constructing the call stack. Although the crash appears only in release, we skip the top call frame for both debug and release in order to keep having the same behaviour in both configurations.


ivanbuhov:
run ci

ivanbuhov:
Closed in favour of https://github.com/NativeScript/ios-runtime/pull/691

PR #702, opened 2016-12-19
Call original domain dispatcher methods when there is no custom implementation

KristinaKoeva:
Chrome dev tools calls Runtime.compile method which does not exist on the current webkit Runtime agent. So instead of moving the Runtime agent to our side, enable having one native and one javascript implementation

Merge with: https://github.com/NativeScript/webkit/pull/15


KristinaKoeva:
run ci

vchimev:
run ci

KristinaKoeva:
run ci

Issue #717, opened 2017-01-20
Should multiple runtimes on the main thread work with the debugger ?


jeduden:
I am currently trying figure out if the ios runtime code supports using multiple runtime instances on the main thread. it looks like the execution of the JS works. However, the debugger doesn't break on break-points in this scenario. 

Also, saw the discussions around webworker. Would the debugger be more stable if one of the runtimes on a different (non-ui) thread ?

ivanbuhov:
@jeduden I am interested in what are you trying to achieve by running two runtime instances on the main thread. What is the problem you are trying to solve? How do you initialize the runtimes and which runtime instance do you pass to the  [`TNSEnableRemoteInspector(argc, argv, runtime);`](https://github.com/NativeScript/ios-runtime/blob/1eea4fa472c6f959f63036fb051b163e22b14d37/cmake/main.m#L22) method call?
We have never had an use case to run more than one runtime on the main thread so this scenario is never tested and we haven't actively tried to support it. However, since I can't think of some technical blocker, from technical perspective it should be feasible. Regarding the debugger, currently only one runtime can be debugged and this is the one passed to [`TNSEnableRemoteInspector(argc, argv, runtime);`](https://github.com/NativeScript/ios-runtime/blob/1eea4fa472c6f959f63036fb051b163e22b14d37/cmake/main.m#L22) method. It is possible to have more than one inspector connected to different debugging frontends and listening on different ports but this requires some effort in the runtime(s) side.
P.S. Currently debugging of worker threads is not supported but you still can debug the runtime running on the main thread.

jeduden:
My use case: i have two different systems driven by javascript code. Conceptually, they are disconnected and in an alternative implementation I naturally used two separate JavaScriptCore instances. 

Since the systems are separate, the event when a reload of the code could happen is also different. I implemented the reload by just throwing away the vm and instantiating a new one.
If I serve both systems from the same vm, i need to take a better look at hot module replacement.

A bit more context why I am asking: 
I stumbled across a section in the code that uses a mapping between thread and vm.
But I didn't see any explicit docs about this topic:
- https://github.com/NativeScript/ios-runtime/blob/9f38cae76f598ab761066d9e139104f2a32b1035/src/NativeScript/TNSRuntime.mm#L50

If there is no technical blocker by design, I can continue with my current approach and see what I find.

ivanbuhov:
[...]

PR #375, opened 2015-10-15
Inspector frontend glue code

KristinaKoeva:
#335

PR #724, opened 2017-02-08
Raise fatal error when inspector_modules.js cannot be loaded.

jeduden:
When the debugger is connected and the module cannot be loaded the debugger does not work. Without reporting the error is can get very difficult to troubleshoot the issue.

### DIFF
--- a/src/NativeScript/TNSRuntime+Inspector.mm
+++ b/src/NativeScript/TNSRuntime+Inspector.mm
@@ -12,6 +12,8 @@
 #include <JavaScriptCore/InspectorAgentBase.h>
 #include <JavaScriptCore/InspectorFrontendChannel.h>
 #include <JavaScriptCore/JSONObject.h>
+#include <JavaScriptCore/JSNativeStdFunction.h>
+#include <JavaScriptCore/JSInternalPromise.h>
 
 #include "GlobalObjectConsoleClient.h"
 #include "GlobalObjectInspectorController.h"
@@ -107,10 +109,26 @@ private:
 
     globalObject->microtasks().swap(other);
 
-    loadAndEvaluateModule(_runtime->_globalObject->globalExec(), WTF::ASCIILiteral("inspector_modules.js"));
-
+    JSInternalPromise* promise = loadAndEvaluateModule(_runtime->_globalObject->globalExec(), WTF::ASCIILiteral("inspector_modules.js"));
+    
+    JSValue error;
+    JSFunction* errorHandler = JSNativeStdFunction::create(*_runtime->_vm.get(), globalObject, 1, String(), [&](ExecState* execState) {
+        error = execState->argument(0);
+        return JSValue::encode(jsUndefined());
+    });
+    promise->then(globalObject->globalExec(), nullptr, errorHandler);
+    
     globalObject->drainMicrotasks();
     globalObject->microtasks().swap(other);
+    
+    if (error) {
+        Exception* exception = jsDynamicCast<Exception*>(error);
+        if (!exception) {
+            exception = Exception::create(*_runtime->_vm.get(), error, Exception::DoNotCaptureStack);
+        }
+        
+        reportFatalErrorBeforeShutdown(globalObject->globalExec(), exception);
+    }
 }
 
 - (void)dispatchMessage:(NSString*)message {


KristinaKoeva:
Hey @jeduden, thank you very much for your contribution !
Would you please explain what are the cases when the `inspector_modules` fails to load ? Do you have any custom modifications in it ?

jeduden:
@KristinaKoeva  this patch is a by-product of my research related to issue #717 . 

TLDR:
When having a non standard js delivery system and/or when one is not conscious of the fact that the debugger needs a couple of js files from the tns_modules in order to work. It is quiet a journey to troubleshoot the issue, since i had to understand the cmake build files and the internals of how the debugger and the runtime work hand in hand in order to diagnose the issue.

With this PR i hope to short-circuit the journey and to report directly back that the debugger is not functional if this error occurs.

Full version: 
The ios nativescript runtime is integrated in an existing application, which has it's own systems to download js.
I was looking into why the js debugger didn't work with this application. The first step was to simplify the scenario and have only one runtime but still the custom js downloader, in that simplified scenario debugging also didn't work. 

The backend could connect and i was able to see source code but breakpoints and so on didn't work. There wasn't any feedback in the logs why the debugger didn't work.

I  build a debuggable version of the ios runtime to see what exactly failed. That step was quiet tricky, since the build scripts seem to support only release builds out of the box.

After i was able to step through the native code, i was able to see that the debugger code tried to load inspector_modules.js which i didn't expect.  But stepping over the load call, i didn't see any  obvious issues. Only later i realized that the call actually returns a promise which needs interpretation before figuring out if load actually worked.

And when i add in the error handler i was finally able to diagnose that the inspector modules were not found. And when I fixed delivery of the tns_modules to app folder i was able to confirm that the inspector modules are loading and debugger started to work as well.

KristinaKoeva:
The thing that isn't clear to me is why the debugger does not work. I've created a simple {N} project and tried the following scenarios:
1. Added `throw "Error"` on the first line of the `inspector_modules` file
1. Made a syntax error in the `inspector_modules` file
1. Deleted the `inspector_modules` file

Neither of them result in a broken debugger experience. 
I completely agree that we should provide logging but since this file is only used for enhancing the debugger experience I don't think that we should kill the user application when something is wrong.

Would you please share something simple that demonstrates the problem ?

jeduden:
@KristinaKoeva  
After reading your email, I was looking into reproducing the issue with a minimal project.

1. In this tag the debugger works: https://github.com/jeduden/min-app-inspector-not-working/releases/tag/debugger-works-out-the-box

Confirming that the inspector modules are not required. So I continued looking why it didn't work for me.

2. In this tag the debugger doesnt work: https://github.com/jeduden/min-app-inspector-not-working/releases/tag/debugger-doesnt-workx
    it uses JSEvaluateString to execute a script. not executeModule

3. In this tag the debugger works again: https://github.com/jeduden/min-app-inspector-not-working/releases/tag/debugger-works-with-tns_modules-again
    it includes the tns_modules in the folder. and the debugger works again.

4. in this tag https://github.com/jeduden/min-app-inspector-not-working/releases/tag/debugger-works-after-calling-exec-module-once
    i tested the hypothesis that the debugger starts working after the first call to executeModule.
   which it does.

I guess loadAndEvaluateModule has a side-effect that initialises/fixes some state essential for the debugger to work, but I am not sure yet why.

KristinaKoeva:
Hey @jeduden,
I had some troubles running the project that you've attached under the second point - https://github.com/jeduden/min-app-inspector-not-working/releases/tag/debugger-doesnt-work
So I've created a new plain {N} project, I've deleted the contents of the app folder and I've replaced the content of main.m in the new project with the content of main.m from the project that you've attached. Still everything works on my side, the debugger waits at the `debugger;` statement and I am able to attach without problems. 

If you don't mind I will close this PR and we will revisit it once we decide whether we should terminate the application or log a warning when there are errors in the `inspector_modules.js` file. You can open an issue where we could try to find what's causing your debugger problems. Is that OK for you ?

jeduden:
@KristinaKoeva sure. no worries, reproducing this is tricky. indeed. 

One question regarding inspector_modules.js if the module is "optional" why is it loaded in the first place ? What functionality does it provide ?

KristinaKoeva:
Some time ago we've implemented [a way to extend the debugger through JavaScript code](https://github.com/NativeScript/ios-runtime/pull/375). The main reason for this was that the runtime is not aware of the UI aspects of the application hence the debugger cannot provide CSS or DOM debugging. The CSS and DOM information is part of the {N} modules that are all written in JavaScript. So we've decided to introduce the `inspector_modules.js` file that should be loaded before everything else in the application. In this file the {N} modules are registering specific handlers that would eventually extend the debugger with CSS or DOM debugging. Currently this file contains only an extension for debugging Network request, but we plan to add more extensions in the future. Here is the code of the [`inspector_modules.js` file](https://github.com/NativeScript/NativeScript/blob/master/tns-core-modules/inspector_modules.ios.ts) This of course would not be available if you don't use {N} modules in your application.

PR #753, opened 2017-04-21
ObjC TS decorators for specifying protocols and exposed methods

PeterStaev:
Implements #754 
Related PRs:
* Core modules: https://github.com/NativeScript/NativeScript/pull/4048
* Docs: https://github.com/NativeScript/ios-runtime-docs/pull/61

### DIFF
--- a/src/NativeScript/inlineFunctions.js
+++ b/src/NativeScript/inlineFunctions.js
@@ -14,6 +14,78 @@ Object.assign(global, {
     NSMakeRange(location, length) {
         return new NSRange({ location, length });
     },
+
+    // Decorators support    
+    __decorate(decorators, target, key, desc) {
+        var c = arguments.length, r = c < 3 ? target : desc === null ? desc = Object.getOwnPropertyDescriptor(target, key) : desc, d;
+        if (typeof Reflect === "object" && typeof Reflect.decorate === "function") r = Reflect.decorate(decorators, target, key, desc);
+        else for (var i = decorators.length - 1; i >= 0; i--) if (d = decorators[i]) r = (c < 3 ? d(r) : c > 3 ? d(target, key, r) : d(target, key)) || r;
+        return c > 3 && r && Object.defineProperty(target, key, r), r;
+    },
+    __param(paramIndex, decorator) {
+        return function (target, key) { decorator(target, key, paramIndex); }
+    },
+
+    // Decorators
+    ObjCClass() {
+        var protocols = Array.from(arguments);
+
+        return function (target) {
+            if (protocols.length > 0) {
+                target.ObjCProtocols = (target.ObjCProtocols && target.ObjCProtocols instanceof Array ? target.ObjCProtocols.concat(protocols) : protocols);
+            }    
+        }
+    },
+    ObjCMethod() {
+        var name = arguments[0];
+        var hasName = (name !== undefined && typeof name === "string");
+        var returnType = (hasName ? arguments[1] : arguments[0]);
+
+        return function (target, propertyKey, descriptor) {
+            if (!target.constructor.ObjCExposedMethods) {
+                target.constructor.ObjCExposedMethods = {};
+            }
+            if (!target.constructor.ObjCExposedMethods[propertyKey]) {
+                target.constructor.ObjCExposedMethods[propertyKey] = {};
+            }
+            target.constructor.ObjCExposedMethods[propertyKey].returns = returnType || interop.types.void;
+
+            if (hasName && name !== propertyKey) {
+                target.constructor.ObjCExposedMethods[name] = target.constructor.ObjCExposedMethods[propertyKey];
+                delete target.constructor.ObjCExposedMethods[propertyKey];
+
+                target[name] = function () { 
+                    return this[propertyKey].apply(this, arguments);
+                }
+            }
+        }
+    },
+    ObjC() {
+        var args = Array.from(arguments);        
+
+        return function (target, propertyKey, descriptor) {
+            if (propertyKey === undefined) {
+                return ObjCClass.apply(this, args)(target);
+            }
+
+            ObjCMethod.apply(this, args)(target, propertyKey, descriptor);
+        };
+    },
+    ObjCParam(type) {
+        return function (target, propertyKey, parameterIndex) {
+            if (!target.constructor.ObjCExposedMethods) {
+                target.constructor.ObjCExposedMethods = {};
+            }
+            if (!target.constructor.ObjCExposedMethods[propertyKey]) {
+                target.constructor.ObjCExposedMethods[propertyKey] = {};
+            }
+            var exposedMethod = target.constructor.ObjCExposedMethods[propertyKey];
+            if (!exposedMethod.params) {
+                exposedMethod.params = [];
+            }
+            exposedMethod.params[parameterIndex] = type || interop.types.void;
+        };
+    },
 });
 
 Object.defineProperty(global, "__tsEnum", {
--- a/build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
+++ b/build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
@@ -242,7 +242,6 @@
 		858B835818CA111C00AB12DE /* Debug */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ALWAYS_SEARCH_USER_PATHS = NO;
 				ARCHS = "$(ARCHS_STANDARD)";
 				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
@@ -280,7 +279,6 @@
 		858B835918CA111C00AB12DE /* Release */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ALWAYS_SEARCH_USER_PATHS = NO;
 				ARCHS = "$(ARCHS_STANDARD)";
 				CLANG_CXX_LANGUAGE_STANDARD = "gnu++0x";
@@ -317,7 +315,6 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = CDD59A261BB43B5D00EC2671 /* build-debug.xcconfig */;
 			buildSettings = {
-				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ARCHS = "$(ARCHS_STANDARD)";
 				CLANG_ENABLE_MODULES = NO;
 				CLANG_MODULES_AUTOLINK = NO;
@@ -340,7 +337,6 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = CDD59A271BB43B5D00EC2671 /* build-release.xcconfig */;
 			buildSettings = {
-				ASSETCATALOG_COMPILER_APPICON_NAME = AppIcon;
 				ARCHS = "$(ARCHS_STANDARD)";
 				CLANG_ENABLE_MODULES = NO;
 				CLANG_MODULES_AUTOLINK = NO;
--- a/tests/TestRunner/app/Inheritance/TypeScriptTests.js
+++ b/tests/TestRunner/app/Inheritance/TypeScriptTests.js
@@ -1,8 +1,7 @@
-// TODO: Use TypeScript definitions when they get ready
 var TSObject = (function (_super) {
     __extends(TSObject, _super);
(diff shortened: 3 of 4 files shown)


src/NativeScript/inlineFunctions.js
```diff
@@ -14,6 +14,119 @@ Object.assign(global, {
...
+
+    //     return function (target, propertyKey, descriptor) {
+    //         if (propertyKey === undefined) { // Class decorator
+    //             var implementation = {};
+    //             for (var key in target.prototype) {
+    //                 if (target.prototype.hasOwnProperty(key)) {
+    //                     implementation[key] = target[key];
+    //                 }    
+    //             }
+    //             var extended = target.extend(implementation, {
```
PanayotCankov: Delete the commented code.
Or use:
```
const baseClass = Object.getPrototypeOf(target.prototype).constructor;
```

PanayotCankov: @ivanbuhov noted the `Object.getPrototypeOf(target)` will work too. The modern JavaScript and the runtime wires the prototype chain of the constructor functions properly, not only the prototype chain of the prototypes objects. This is so cool. :+1:

PeterStaev:
Refactored to `@ObjC` unified decorator for classes/methods. Couple of notes:
1. In order for the decorators to work, even if you do not implement protocols, you are required to decorate the class with `@ObjC`
2. Because of the current TS extend which also tries to register the class, a warning is output that the class is already used.

src/NativeScript/inlineFunctions.js
```diff
@@ -14,6 +14,103 @@ Object.assign(global, {
...
+            // Method Decorator
+            if (!target.constructor.ObjCExposedMethods) {
+                target.constructor.ObjCExposedMethods = {};
+            }
+            if (!target.constructor.ObjCExposedMethods[propertyKey]) {
+                target.constructor.ObjCExposedMethods[propertyKey] = {};
+            }
+            target.constructor.ObjCExposedMethods[propertyKey].returns = additionalParams[0] || interop.types.void;
+
+            if (hasName && name !== propertyKey) {
```
ivanbuhov: @PeterStaev As far as I understand, the idea behind this is to have ObjC friendly selector and JS-friendly method name simultaneously. Keep in mind that having a JS-unfriendly name is a concern only if we want to actually call the method from JS (often the exposed methods are only used in ObjC) and even without implicitly adding additional method you are still able to achieve the desired effect by adding it explicitly:
```
...
ObjCMethod(...)
public "selector:With:Parameters:" (x, y, z) {}

public selectorWithParameters() {
    return this["selector:With:Parameters:"].apply(this, arguments);
}
...
```
The explicit approach is more verbose (which is a con) but
1. It will eliminate the need of doing yet another special thing under the hood.
2. The method with the JS-friendly name may already exists in the prototype object.

src/NativeScript/inlineFunctions.js
```diff
@@ -14,6 +14,103 @@ Object.assign(global, {
     NSMakeRange(location, length) {
         return new NSRange({ location, length });
     },
+
+    // Decorators support    
+    __decorate(decorators, target, key, desc) {
```
PeterStaev: Yes, they are the same, but with formatting and also changed `Reflect` to be `global.Reflect` as I saw this is what you use in the android runtime. I will change to follow the formatting of the TS default ones.

src/NativeScript/inlineFunctions.js
```diff
@@ -14,6 +14,103 @@ Object.assign(global, {
...
+                var baseTarget = Object.getPrototypeOf(target);
+                
+                for (var prop in target.prototype) {
+                    if (target.prototype.hasOwnProperty(prop)) {
+                        implementation[prop] = target.prototype[prop];
+                    }
+                }
+
+                var newName = hasName ? name : target.name;                
+                var extended = baseTarget.extend(implementation, {
```
PeterStaev: The benefit of using the `baseTarget.extend` is that we can specify the `name` of the class, which with the TS extends method is impossible. Also it will allow the TS extending to match and use the JS extend, so if any changes are made to the JS extend they can easily flow over to TS. Now the problem is that because of the TS extends, this creates two instances of the class. 

You can take a look at the suggestion by @PanayotCankov here: https://github.com/NativeScript/NativeScript/issues/4018#issuecomment-295600104 . Having the aggregated `@ObjC` decorator looked like a good idea initially, but last night when I tried to make the definitions, it was impossible to make them strongly typed. So I think now having different methods is the better option. But IMO the `ObjCMethod` one should just accept the things concerning the method return type/name, and not accept a JSON with the parameters and everything. Also this creates some confusion to the end user. They will need to know if the param decorator or the method signature takes precedence.

src/NativeScript/inlineFunctions.js
PanayotCankov: I don't think either point is valid and here is why:
1. The decorators provide TypeScript friendly API over existing functionality or provide shorthand for frequently used patterns. I needed the special thing and as a TypeScript developer I think it is convenient
2. As a TypeScript developer I will probably check the methods of the base class when I am picking the TypeScript friendly name, just like I would when I am adding any other method that may unintentionally override a base class method.

src/NativeScript/inlineFunctions.js
PanayotCankov: :+1: The runtime really does a good work in __extends so calling .extend was my mistake. I am so sorry. That's the reason I wanted someone with better understanding in the runtimes to check this and help in its development.

On the other hand method overloading in TypeScript is trivial and done on a daily basis. The class, method and parameter decorators have really distinct signatures. If you are worried by the overload selection I could recommend you to just refactor the selection a little and separate the overload selection from the implementation similar to this:
https://github.com/pana-cc/mocha-typescript/blob/master/index.ts#L308
https://github.com/pana-cc/mocha-typescript/blob/master/index.ts#L329

src/NativeScript/inlineFunctions.js
PeterStaev: Yes, this is exactly the purpose of this. True that it is a bit under the hood. For the JS-friendly name being existent in the proto, I think it is not that big of a problem as you can always make a unique name that will work. I know you can call the method even with the current implementation of the exposed methods. But if you look at the above explicit approach, the code is more or less copy paste with making the same adjustments to every snippet - add a function with the colons removed and call `this` with the colon delimited name and sending the arguments. If you have to do this 100 times, it will not be pretty 😄 But then again, so far, I have never needed to call the exposed method from JS 😉 So I'm ok with both approaaches.

PeterStaev:
@ivanbuhov @PanayotCankov I have no problems with rebasing this onto the release branch. But let's settle an API definition and I will see to make the necessary changes.

src/NativeScript/inlineFunctions.js
ivanbuhov: > But IMO the ObjCMethod one should just accept the things concerning the method return type/name, and not accept a JSON with the parameters and everything. Also this creates some confusion to the end user. They will need to know if the param decorator or the method signature takes precedence.

Makes sense. I agree.

 > The benefit of using the baseTarget.extend is that we can specify the name of the class, which with the TS extends method is impossible.

We are setting the name of the native class to be the same as the name of the TS class and you can name your TS class however you want, so you have full control over the name of the native class. Still creating two classes when only one is needed is not only unnecessary but it possibly has a impact on the performance. However, if we really want to specify the name of the native class via a decorator (if we want the name not to be the same as the TS class), I suggest changing our TS extend to support it instead of calling `extend` twice.

P.S. Just want to add that it will be good if the decorator doesn't override the the `ObjCProtocols` static property but extend it. This way you can set some of your protocols  as a static property and some of them in a decorator.

src/NativeScript/inlineFunctions.js
ivanbuhov: @PanayotCankov Makes sense, I agree. Lets keep the name parameter.

src/NativeScript/inlineFunctions.js
PanayotCankov: I too agree that it would be better to avoid .extends if we can. Making two separate derived classes is not worth the benefit of specifying the native name in a decorator. For Java the native name also has package and the name in an Android decorator makes much more sense, but for iOS, given the iOS class names can easily be expressed in TypeScript, I don't see that much benefit now.

PeterStaev: The benefit I see is when JS is minified. Right now we have to maintain a long list of entities that are exluded from mangling. It would be good if the TS extend is changed and looks at the 'name' property of the class. This way we can specify the name here w/o using base extend.

ivanbuhov: I see the benefit of specifying the name in the decorator and I am not against it but I don't agree with the current implementation. I suggest refactoring the [`ObjCClassBuilder`](https://github.com/NativeScript/ios-runtime/blob/master/src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.h) to not require the name of the native class so early but this may be a bit tricky and is a good candidate for another PR.

PeterStaev: @ivanbuhov  Agreed. 

@PanayotCankov , for overloading, the problem is not with the implementation of it. It is already working fine. The problem is with the definitions, and mainly comes because the `returnType` for the method and protocols optional parameters are any. So in the d.ts the definitions conflict with each other. For example:
```ts
function ObjC(returnType: any): MethodDecorator;
function ObjC(...protocols: any[]): ClassDecorator;
```
These will not work. You have to define them as:
```ts
function ObjC(type: any): ClassDecorator & MethodDecorator;
function ObjC(...protocols: any[]): ClassDecorator;
```
But then the help provided in the tooltip will not be very clear because `type` for the class will actually be a protocol and for the method this will be the return type.

PanayotCankov: @PeterStaev I'd prefer single ObjC as I'd now what it does, but you are right that using separate functions will allow for stricter typing and better docs. With so much any types you can't limit what can be provided as arguments to these function, so at the moment except the doc for each separate function the type checking won't benefit much. Either way I would be happy, its up to you to decide.

ivanbuhov:
@PeterStaev @PanayotCankov  Guys, taking all arguments into account, I came up with summarized  spec to discuss and (not) agree on.
1. Class decorator factory named `ObjC` with the following signature:
```
function ObjC(...protocols: any[]): ClassDecorator;
```
Used like this:
```
ObjC(ProtocolA, ProtocolB)
class A extends B implements ProtocolA, ProtocolB {}
```
It should not call the `extend` function of the base class. It should add more protocols to the static `ObjCProtocols` array **without overriding** it.

Example implementation:
```
ObjC() {
    return function (target) {
        target.ObjCProtocols = (target.ObjCProtocols || []).concat(arguments);
     };
}
```

2. Method Decorator factory named `ObjCMethod` - it should set the method's returned value type in the `ObjCExposedMethods` static property. It will also accept an ObjC-friendly selector name by which the method will be exposed in ObjC, preserving the JS-friendly name in case the method is called from JS. Since both parameters are not required (the returned type default value is `interop.type.void` and the name default value is the JS-friendly name of the method) we have to figure out how to pass both parameters. I am a fan of the more verbose (but explicit) option - `ObjC({returns: foo, selector: "foo:bar:"})` because it will eliminate the need to type-check the arguments in order to guess which of the supported signature is used, but it is not a strong opinion and if you don't like it we can come up with other solution.
3. `ObjCParam(type)` - I think we all agree that the current implementation provided by @PeterStaev is fine.

@PanayotCankov @PeterStaev I would like to hear your opinions.

PeterStaev:
For most parts I agree, with the following comments:
1. If we are going to have separate decorators isn't it better to follow a common naming pattern? So the class decorator to be named `ObjCClass`, envisioning a future add of a `name` parameter? 
2. Personally I prefer to have separate params and not as JSON. Easier to type and I do not think there will be other parameters being added to the method decorator.

PeterStaev:
Updated the PR:
1. `ObjCClass` - decorator for classes, accepting protocols as arguments. Extends the existing protocols (if any)
2. `ObjCMethod` - method decorator, with optional name and return type arguments with type check. 
3. `ObjC` - shorthand decorator for either classes or methods that calls the above ones
4. Replaced the `__decorate` and `__param` with the unformatted ones. 
5. **Rebased to the release branch**. 

I think this should make everyone happy 😄 

Once this is approved I will see to adjust the docs/core PRs.

ivanbuhov:
@PeterStaev  Great job! Lets squash the changes before merging it in order to keep the history clean. :)

PeterStaev:
Done. Also updated the other 2 PRs.

PanayotCankov:
Decorate all the classes

jlooper:
@PeterStaev This PR qualified for special treatment! Congratulations on your new badge! 

contrib

PR #766, opened 2017-05-15
Add a high resolution __time on the global object

PanayotCankov:
Calling
```
var n = __time();
``` 
Will get the current time in milliseconds but it is accuracy goes beyond millis. About 3 digit micro secs. are stored as fraction. Date.now() truncates to milis and is about 6 times slower.


ivanbuhov:
run ci

ivanbuhov:
run ci

PanayotCankov:
run-ci

ivanbuhov:
run ci

PR #771, opened 2017-06-13
Add manual instrumentation API and profile key runtime points

PanayotCankov:


src/NativeScript/GlobalObject.moduleLoader.mm
```diff
@@ -390,6 +393,15 @@
                                                   args.append(moduleKey);
                                                   JSValue entry = JSC::call(execState, function, callType, callData, moduleLoader, args);
                                                   record = jsCast<JSModuleRecord*>(entry.get(execState, Identifier::fromString(execState, "module")));
+                          
+                                                  if (frame.check()) {
```
ivanbuhov: If I have understood the API usage correctly, instrumenting a piece of code requires the following changes:
 * Create a frame (`tns::instrumentation::Frame frame;`) just before the code that needs to be instrumented.
 * At the end of the code to be profiled we check if 16000 microseconds have passed since the frame creation (I can't see why this is needed?) and only then we have some logic specific to the event we are profiling, in most cases ending with logging. 

What bothers me is that doing so, we are mixing the profiling-related logic with the actual code and if one day we are profiling a more significant number of events, this may lead to higher maintenance cost. Ideally, we should have a one-liner at the end of the code to be profiled. I understand that the ending logic can't be extracted in a single method because it is different depending on the code/event we are profiling (we are logging different kind of event-specific information), but still we can have event specific methods: `endOfMyCustomEvent1WithFrame::withCustomParam:`, `endOfEvent2withFrame:param2:param3:` etc.

src/NativeScript/ManualInstrumentation.h
```diff
@@ -0,0 +1,77 @@
...
+                }
+            }
+            
+            inline bool check() const {
+                if (mode == Mode::Disabled) {
+                    return false;
+                }
+                std::chrono::steady_clock::time_point end = std::chrono::steady_clock::now();
+                auto duration = std::chrono::duration_cast<std::chrono::microseconds>(std::chrono::operator-(end, start)).count();
+                return duration >= 16000;
```
PanayotCankov: It means we are not interested in fast frames and don't want to large amount of small frames. Logging may be slow.

src/NativeScript/GlobalObject.moduleLoader.mm
PanayotCankov: There is one-liner `TNSPERF()` that does `tns::instrumentation::Frame frame(__func__)` that will print the method name if the method takes too much time, this better be placed the the start of a method and relies on the Frame C++ destructor to mark the end of the method.

The if-ed stuff is done to minimise runtime costs when running without profiling. Most of the if-ed logic in the places where we are mixing logic is stuff that will extract string and combine them, that is cost that we won't pay if the profiling is not enabled or if the frame was actually fast (<16ms.).

I have no idea what you may need one day. I'd be glad for further contribution on this.

`checkForMyCustomEvent1WithFrame::withCustomParam:` and the rest will make the profile know about random stuff in the runtime, I don't think the Frame should be coupled with anything else but timing and logging.

PanayotCankov:
Instruments won't add the extra information (for example if require is slow, which module was required or if materialise properties is slow - which class was materialised) and in general can't be easily mixed with JS (I am not sure about that) so WebInspector is great for JS and the Instruments for native code, but I haven't found a tool to do properly both.

PR #780, opened 2017-07-27
Update LLVM/Clang to v4.0.0

ivanbuhov:
To update to llvm 3.8:

1. Download the [pre-build binary of LLVM 4.0.0 with Clang for Mac OS X](http://releases.llvm.org/download.html#4.0.0).
2. Extract the content of the package (/usr/local/lib is a good candidate for a destination folder)
3. Include the path to `{llvm-root}/bin/llvm-config` or symlink to it in your `PATH`
4. Reset the cmake cache of the metadata generator project. This way the generator will be linked with the new version of Clang/LLVM.


mm-ns:
Hi Ivan, I am the person that reported this to Sebastian, so I appreciate your support on this. What would you say the time estimate on this is?

PR #795, opened 2017-09-13
Add caching for function and method ffi_cifs

tdermendjiev:
As unifying ffi signatures in metadata-generator would lead to insufficient ffi metadata in ios-runtime only ffi_cif caching in FFICall initialization is implemented.


src/NativeScript/Calling/FFICache.h
```diff
@@ -0,0 +1,45 @@
...
+    
+    struct SignatureHash {
+        
+        std::size_t operator()(std::vector<const ffi_type*> signature) const {
+            std::size_t seed = 0;
+            
+            
+            for(size_t i = 0; i < signature.size(); i++) {
+                seed <<= 4;
+                seed |= signature[i]->type;
```
mbektchiev: I'd recommend using the [Fowler–Noll–Vo hash function](https://en.wikipedia.org/wiki/Fowler%E2%80%93Noll%E2%80%93Vo_hash_function). This one seems to not be distributive enough.

src/NativeScript/Calling/FFICall.cpp
```diff
@@ -57,17 +64,34 @@ void FFICall::initializeFFI(VM& vm, const InvocationHooks& hooks, JSCell* return
...
         this->_stackSize += malloc_good_size(std::max(this->_cif->arg_types[i]->size, sizeof(ffi_arg)));
     }
 }
 
+std::shared_ptr<ffi_cif> FFICall::checkForExistingCif(unsigned int nargs, ffi_type* rtype, ffi_type** atypes) {
+
+    std::unordered_map<std::vector<const ffi_type*>, std::shared_ptr<ffi_cif>, SignatureHash>::const_iterator it = FFICache::global()->cifCache.find(this->signatureVector);
+
+    if (it == FFICache::global()->cifCache.end()) {
+        std::shared_ptr<ffi_cif> shared(new ffi_cif, deleteCif);
```
ivanbuhov: I find it a bit confusing that a method called `checkForExistingCif` actually creates a new `cif` in case there is no existing one. I suggest renaming the method (e.g. `getCif` or something) or moving the `cif` creation logic outside the method.
However, this is more or less a subjective statement, so if you don't agree, I am perfectly fine to leave it as is.

src/NativeScript/Calling/FFICache.h
```diff
@@ -0,0 +1,39 @@
...
+            seed = (seed ^ reinterpret_cast<size_t>(signature[i])) * 16777619U;
+        }
+        return seed;
+    }
+};
+
+class FFICache {
+
+public:
+    std::unordered_map<std::vector<const ffi_type*>, std::shared_ptr<ffi_cif>, SignatureHash> cifCache;
```
mbektchiev: I suggest you declare a new typedef in order to avoid repeating the whole `std::unordered_map<std::vector<const ffi_type*>, std::shared_ptr<ffi_cif>, SignatureHash>` everywhere you need it.

PR #799, opened 2017-09-28
Debugging from Xcode documentation and improvements

mbektchiev:
* Add `launch-proxy.sh` which npm installs and lauches the proxy in a `while true` block
* Add package.json for dependencies of proxy
* Add instructions for debugging from Xcode to README.md

### DIFF
--- a/src/debugging/README.md
+++ b/src/debugging/README.md
@@ -1,5 +1,30 @@
-To run in Google Chrome:
+# Attach JS Debugger on Run from Xcode
+1. `cd src/debugging`
+2. `./launch-proxy.sh`
+3. In Xcode add `--nativescript-debug-brk` command line arg in the app scheme of the NativeScript application to be debugged
+4. Launch the application with Xcode and (within the 30 seconds timeout for the frontend to attach) open one of the following:
+
+
+### In {N} inspector application:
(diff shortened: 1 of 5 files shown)


src/debugging/README.md
```diff
@@ -1,4 +1,11 @@
-To run in Google Chrome:
+# Attach on Run from Xcode
+1. `cd src/debugging`
+1. `./launch-proxy.sh`
+1. In Xcode add `--nativescript-debug-brk` command line arg in the app scheme of the NativeScript application to be debugged
+1. Open WebInspectorUI/Main.html in latest Safari version
```
KristinaKoeva: The latest Safari version may not be the right one. When you update JavaScriptCore you should also update the debugging frontend which may require features that are included only in the nightly Safari version that uses that specific JavaScriptCore and are missing from the official one :)

mbektchiev: Thanks for the clarification, @KristinaKoeva :) I've added the following paragraph below. Does it sound better this way?

###### Note that if you've just updated the WebKit version even the canary version of Safari may be missing some features used by the frontend. The latest nightly Safari version that uses that specific JavaScriptCore must be used in this case. Updating and re-building `src/debugging/Inspector/Inspector` application should always work as well.

PR #803, opened 2017-10-10
Pass compiler identification variables to WebKit CMake

mbektchiev:
Xcode 9 fails to build the test project because it is 32-bit but
iOS 11 no longer supports 32-bit binaries


dtopuzov:
run ci

mbektchiev:
run ci

PR #812, opened 2017-10-25
Load inspector pages modules deterministically before launching app's main module

mbektchiev:
Loading them asynchronously on inspector client connection can sporadically break
application startup due to the common dependencies between the application and
inspector modules.


mbektchiev:
This fix is wrong and is reverted with https://github.com/NativeScript/ios-runtime/pull/813

The inspector modules must be executed after
the frontend client has connected because the domain handlers registration happens
through calling the `__inspectorSendEvent` function which sends a JSON message to the
frontend inspector socket.

Issue #807, opened 2017-10-20
Feature Request: Better error logging for the built in __extends


NathanaelA:
@NathanWalker and I were debugging an issue that only occurred on the iOS runtimes.  
It would give the awesome error `b.prototype` error from `__extends`
```
1   0x114ee7bf8 -[TNSRuntime executeModule:referredBy:]
2   0x10f6231e1 main
3   0x116c1665d start
4   0x1
undefined:7:68: JS ERROR TypeError: undefined is not an object (evaluating 'b.prototype')
Oct 19 17:11:01 NathanMac com.apple.CoreSimulator.SimDevice.3A54BCF4-85F7-4B94-8D6F-9C006FE91F2D[22248] (UIKitApplication:com.ugroupmedia.pnp.client.mobile[0xabfd][22270][27845]): Service exited due to Segmentation fault: 11
```

I created a custom iOS runtime with a modified __extends that looked like this
```
(function (d, b) {
try {
    for (var p in b) if (b.hasOwnProperty(p)) d[p] = b[p];
    function __() { this.constructor = d; }
    d.prototype = b === null ? Object.create(b) : (__.prototype = b.prototype, new __());
} catch(e) { 
      console.log("Error:", e, e.stack); 
      throw e;
}
});
```

This allowed us to get a real stack trace.

My suggestion is that in debug mode the engine loads a `__debug_extends.js` file; and in release it loads the normal __extends.js file.  This way we can actually debug it when __extends starts throwing errors.

NathanWalker:
This would have saved myself and the team I’m working with a full 48 hours of misery. Cannot upvote this enough. Thank you @NathanaelA again for the help with the custom runtime to allow us to get past the error which ended up being related to nativescript-angular/animations bundle.

fealebenpae:
I believe TypeScript allows `__extends` to be overridden entirely in JavaScript by setting it on the global object before any transpiled code is loaded.

NathanWalker:
We tried that but it didn’t work for nativescript Angular projects specifically.

NathanaelA:
@fealebenpae - https://github.com/NativeScript/ios-runtime/blob/c949250601ca3072687512225da541f712eae903/src/NativeScript/GlobalObject.mm#L199

It is marked as ReadOnly.   My understanding is that you DO want this readonly so that things can't change it; because in the past we had issues where the internal was overwritten by an external __extends and that caused issues.

ginev:
[...]

Issue #565, opened 2016-05-20
Missing stack trace with async mocha tests


jasssonpet:
Reproduced with NativeScript 2.0.
## Execute the following commands:

``` shell
$ tns create TestApp
$ tns test init
$ tns test ios --emulator
```
## Add the following test:

``` javascript
describe('Test', function () {
    it('Test', function (done) {
        setTimeout(function() {
            assert.equal(-1, 1);
            done();
        }, 1000);
    });
});
```
## The output is:

```
1   0xe0050 NativeScript::FFICallback<NativeScript::ObjCMethodCallback>::ffiClosureCallback(ffi_cif*, void*, void**, void*)
2   0x7a7c50 ffi_closure_inner
3   0x7a85a2 ffi_closure_i386
4   0x380c136 __CFRUNLOOP_IS_CALLING_OUT_TO_A_TIMER_CALLBACK_FUNCTION__
5   0x380bb32 __CFRunLoopDoTimer
6   0x37ca39a __CFRunLoopRun
7   0x37c9846 CFRunLoopRunSpecific
8   0x37c965b CFRunLoopRunInMode
9   0x5573664 GSEventRunModal
10  0x55734a1 GSEventRun
11  0x1568eb9 UIApplicationMain
12  0x7a8468 ffi_call_i386
undefined: JS ERROR AssertionError: expected -1 to equal 1
```
## Expected

The JavaScript stack trace should be printed along with the `sourceURL` for the `AssertionError`.

Ping @ginev

jasssonpet:
This issue can be reproduced with:

```javascript
new Function(`throw new Error("Test error")`)()
```

from the [nativescript-unit-test-runner](https://github.com/NativeScript/nativescript-unit-test-runner/blob/f735a656b70a834bbc5e2c441f87dd9d42062e14/main-view-model.ts#L261).

PR #815, opened 2017-10-25
When an error is thrown in a chained native promise the printed stack is insufficient

ginev:
Discussion: #807

### DIFF
--- a/src/NativeScript/JSErrors.mm
+++ b/src/NativeScript/JSErrors.mm
@@ -63,7 +63,7 @@ void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception, 
         std::cerr << "Native stack trace:\n";
         WTFReportBacktrace();
 
-        std::cerr << "JavaScript stack trace:\n";
+        std::cerr << "\nJavaScript stack trace:\n";
         RefPtr<Inspector::ScriptCallStack> callStack = Inspector::createScriptCallStackFromException(execState, exception, Inspector::ScriptCallStack::maxCallStackSizeToCapture);
         for (size_t i = 0; i < callStack->size(); ++i) {
             Inspector::ScriptCallFrame frame = callStack->at(i);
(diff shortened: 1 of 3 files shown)


src/NativeScript/JSErrors.mm
```diff
@@ -63,7 +63,7 @@ void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception,
         std::cerr << "Native stack trace:\n";
         WTFReportBacktrace();
 
-        std::cerr << "JavaScript stack trace:\n";
```
mbektchiev: We'd better use `NSLog` instead of `std::cerr` because right now stderr lines are not added to the device logs.

ginev: I tend to agree with that but we need to see if we want to swap all `std:cerr` calls with NSLog.

src/NativeScript/JSErrors.mm
```diff
@@ -73,6 +73,34 @@ void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception,
             }
             std::cerr << "\n";
         }
+        
+        //OBSERVATION:
```
zahhak: I would like to see this observation validated.
We can leave it as technical debt (although I really don't like this option) but we must know what is the reason for this behavior in JSC.
In my mind this is an issue that was there for at least an year. It can't be so critical that we would like to merge it immediately without knowing what is the specific reason for this behavior.

ginev: @zahhak - in terms of behaviour the observation is validated. The open question here is whether this is an issue with JSC or an issue with our promise-based implementation for module resolution. I agree that we can spend some more time debugging here and decide if we want to release this as a patch or talk to Apple about it and do it afterwards.

ginev:
There is an update to this fix which includes a change in the JavaScriptCore logic that we rely on:
https://github.com/NativeScript/webkit/pull/18
 Let's continue the discussion: @zahhak, @mbektchiev, @PanayotCankov, @tdermendjiev.

ginev:
This PR addresses https://github.com/NativeScript/ios-runtime/issues/565 as well.

Issue #822, opened 2017-11-10
Incorrrect module path resolution


NickIliev:
Related to [this issue](https://github.com/NativeScript/android-runtime/issues/334) which was resolved in android-runtime. Also logged and discussed [here](https://github.com/NativeScript/NativeScript/issues/3501#issuecomment-343457582)

Steps to reproduce:
- use [this test application](https://github.com/NickIliev/NS-Issues/tree/master/NativeScript/issue_3501_new)
- without any modification, the `moment` library would work on Android and will not work on iOS.
using[ this "hack"](https://github.com/NativeScript/NativeScript/issues/3501#issuecomment-343456812) will also resolve the issue for iOS.

Milestone: Correct the module path resolution for iOS as it works in Android.

etabakov:
Closing this issue as resolved.

PR #824, opened 2017-11-15
Fixes issues with resolution of canonical paths 

ginev:
This PR addresses an issue with absolute module paths not being resolved correctly in certain cases. For example, paths like ./A/moment.js and ../moment.js  did not resolve to the same path and the `moment.js` module was loaded twice like a different module.

See issue #822 for more details.

### DIFF
--- a/src/NativeScript/GlobalObject.moduleLoader.mm
+++ b/src/NativeScript/GlobalObject.moduleLoader.mm
@@ -48,8 +48,6 @@ static NSString* stat(NSString* path) {
 }
 
 static NSString* resolveAbsolutePath(NSString* absolutePath, WTF::HashMap<WTF::String, WTF::String, WTF::ASCIICaseInsensitiveHash>& cache, NSError** error) {
-  
-
     if (cache.contains(absolutePath)) {
         return cache.get(absolutePath);
     }
--- a/src/NativeScript/GlobalObject.moduleLoader.mm
+++ b/src/NativeScript/GlobalObject.moduleLoader.mm
@@ -48,6 +48,8 @@ static NSString* stat(NSString* path) {
 }
 
 static NSString* resolveAbsolutePath(NSString* absolutePath, WTF::HashMap<WTF::String, WTF::String, WTF::ASCIICaseInsensitiveHash>& cache, NSError** error) {
+  
+
     if (cache.contains(absolutePath)) {
         return cache.get(absolutePath);
     }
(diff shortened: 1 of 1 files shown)


src/NativeScript/GlobalObject.moduleLoader.mm
```diff
@@ -154,7 +159,7 @@
             absolutePath = [static_cast<NSString*>(self->applicationPath()) stringByAppendingPathComponent:@"app/tns_modules/tns-core-modules"];
             isModuleRequire = true;
         }
-        absolutePath = [normalizePath([absolutePath stringByAppendingPathComponent:path]) stringByStandardizingPath];
+        absolutePath = normalizePath([absolutePath stringByAppendingPathComponent:path]);
```
mbektchiev: The call to `normalizePath` seems redundant here. In fact I think that the `normalizePath` function is redundant and I don't see it doing anything more than `stringByStandardizingPath` which we always call.

ginev:
@PanayotCankov - we think the `normalizePath` call here: https://github.com/NativeScript/ios-runtime/commit/fba35eab7fe12a307a64aeda06f65b9aaa1e07e5 is redundant. Can you shed some more light on why it was introduced considering that we already call `[NSString stringByStandardizingPath]` on the module path?

PanayotCankov:
https://developer.apple.com/documentation/foundation/nsstring/1407194-stringbystandardizingpath

For absolute paths only, resolving references to the parent directory (that is, the component “..”) to the real parent directory if possible using stringByResolvingSymlinksInPath. For relative paths, references to the parent directory are left in place.

When live-sync-ed without --sync-all-files, the CLI and the runtime work in a way the app folder only is delivered to the device in an external folder, and the node_modules (tns_modules on device) are symlinked, then there are cases where absolute paths that are concatenated with relative paths are resolved to the real directory in the app bundle. Then there are cases where from some modules, content in the app folder is searched, again by relative paths. This ended up to a state where the same module may be executed twice, once in the symlinked path and once in the original (e.g. have two "application" executed leading to funny stuff such as "there is no main entry specified" on startup).

I couldn't find iOS method that will collapse paths such as "/abv/../cde" to "/cde" without resolving symlinks to real paths, and that's why normalizePath is there, we really need it to resolve .. and . in paths without following symlinks.

ginev:
@PanayotCankov - we've introduced a change in TKLiveSync that replaces sym-linking with hard-linking which should allow us to remove the `normalizeString` call and solely rely on [String stringByStandardizingPath] call.

ginev:
Well, it seems hard-linking is not working for us either as the app bundle and the livesync package live in different partitions. We will resort to solely using [String stringByStandardizingPath] at the point where the main token is appended to the absolute path. @mbektchiev @tdermendjiev @PanayotCankov .

PR #827, opened 2017-11-27
Handle EINPROGRESS when connecting to backend socket

mbektchiev:
`connect` is asynchronous and can sometimes return `EINPROGRESS` error code which must be
handled by calling `select` to complete the connection operation.


mbektchiev:
run ci

PR #834, opened 2017-12-07
chore: Fix build with NPM 5

mbektchiev:
`npm pack package` now treats `package` as a npm published package
instead of a local directory name. Disambiguate by changing it to `./package`


mbektchiev:
Approved in https://github.com/NativeScript/ios-runtime/pull/833

PR #835, opened 2017-12-11
fix(inspector):  Fix app crashes on second debugger attach

### LINKED ISSUE #832: Runtime crashes when trying to debug application with --start option
When trying to debug application with the `--start` option the first debug session is OK. If I start the debugger again without restarting the application on the device - the runtime crashes.
After some investigation I found that this call -> https://github.com/NativeScript/ios-runtime/blob/master/src/debugging/TNSDebugging.h#L77 causes the crash. If I remove the error check the debugger will not start but the third time I start debugging with `--start` it works.

Steps to reproduce:
1. `tns create debug-test`
2. `cd debug-test`
3. Deploy the application and start it.
4. `tns debug ios --start --chrome`
5. Open the debug url.
6. After the Chrome DevTools are loaded and the application can be debugged - stop the debug command and close the DevTools tab. Don't restart the application on the device.
7. `tns debug ios --start --chrome`
8. Open the debug url.
9. The application crashes on the device.

tdermendjiev:
# Introduction
We identified a couple of issues with the way connections are handled between the Inspector and the backend. On one side, the sockets were not handled correctly using the Dispatch API resulting in faulty connections upon subsequent connection attempts. 
On the other side, there was an issue with the way Backend Dispatchers were handled via RefPtrs. For example, the DomainBackendDispatcher adopts a pointer to the dispatcher created by the DomainInspectorAgent, which also has a pointer to it. Adopting this pointer does not increase the refcount which in turn breaks the underlying object's (RefCountedBase) state.
The third amendment is in the way backend dispatchers are handled by the Agents in the `didCreateFrontendAndBackend` and `willDestroyFrontendAndBackend` calls. We now create a single instance of each dispatcher and leave it alive until the owning Agent is destroyed. This is the behaviour as seen in the vanilla WebKit [source](https://github.com/WebKit/webkit/blob/0a19ca4ea7d983b71f0caa78cbe13b76050c8022/Source/WebCore/inspector/agents/InspectorNetworkAgent.cpp#L168-L169).

Fixes #832 

* After the debugger frontend is detached certain RefPtr properties are cleaned up and recreated on a subsequent connection which is causing a crash due to improper reference management. Making the *m_backendDispatcher* property persist between different connections fixes the bug.

* Every time the debugger is detached the IO channel should be closed in order the next connection to be successful. Otherwise the connection attempt will fail with a bad descriptor error.

* https://github.com/NativeScript/ios-runtime/pull/835/commits/153d45f01a53f06195968de94dc7bd47ab0cddf2 addresses an issue with debugger delay killing the already existing listen connection at an inappropriate time. We do not need this delay to cancel the listen connection as we immediately close it after a successful frontend connection.

### DIFF
--- a/src/NativeScript/TNSRuntime+Inspector.mm
+++ b/src/NativeScript/TNSRuntime+Inspector.mm
@@ -132,6 +132,10 @@ private:
     self->_inspectorController->pause();
 }
 
+- (bool)hasFrontends {
+    return self->_inspectorController->frontendRouter().hasFrontends();
+}
+
 - (void)dealloc {
     self->_inspectorController->disconnectFrontend(_frontendChannel.get());
     [self->_runtime release];
--- a/src/NativeScript/TNSRuntime+Inspector.mm
+++ b/src/NativeScript/TNSRuntime+Inspector.mm
@@ -57,6 +57,8 @@ private:
 @implementation TNSRuntime (Inspector)
 
 - (TNSRuntimeInspector*)attachInspectorWithHandler:(TNSRuntimeInspectorMessageHandler)messageHandler {
+    JSC::JSLockHolder lock(self->_vm.get());
+
     TNSRuntimeInspector* runtimeInspector = [[TNSRuntimeInspector alloc] initWithRuntime:self
                                                                           messageHandler:messageHandler];
 
@@ -133,10 +135,14 @@ private:
 }
 
 - (bool)hasFrontends {
+    JSC::JSLockHolder lock(_runtime->_vm.get());
+
     return self->_inspectorController->frontendRouter().hasFrontends();
 }
 
 - (void)dealloc {
+    JSC::JSLockHolder lock(_runtime->_vm.get());
+
     self->_inspectorController->disconnectFrontend(_frontendChannel.get());
     [self->_runtime release];
     [super dealloc];
--- a/src/NativeScript/TNSRuntime+Inspector.mm
+++ b/src/NativeScript/TNSRuntime+Inspector.mm
@@ -134,12 +134,6 @@ private:
     self->_inspectorController->pause();
 }
 
-- (bool)hasFrontends {
-    JSC::JSLockHolder lock(_runtime->_vm.get());
-
-    return self->_inspectorController->frontendRouter().hasFrontends();
-}
-
 - (void)dealloc {
     JSC::JSLockHolder lock(_runtime->_vm.get());
 
--- a/src/debugging/TNSDebugging.h
+++ b/src/debugging/TNSDebugging.h
@@ -26,301 +26,353 @@ typedef void (^TNSInspectorProtocolHandler)(NSString *message, NSError *error);
 typedef void (^TNSInspectorSendMessageBlock)(NSString *message);
 
 typedef TNSInspectorProtocolHandler (^TNSInspectorFrontendConnectedHandler)(
-    TNSInspectorSendMessageBlock sendMessageToFrontend, NSError *error);
+                                                                            TNSInspectorSendMessageBlock sendMessageToFrontend, NSError *error);
+
+typedef void (^TNSInspectorIoErrorHandler)(NSObject* dummy/*make compatible with CheckError macro*/, NSError *error);
 
 #define CheckError(retval, handler)                                            \
-  ({                                                                           \
-    int errorCode = (int)retval;                                               \
-    BOOL success = NO;                                                         \
-    if (errorCode == 0)                                                        \
-      success = YES;                                                           \
-    else if (errorCode == -1)                                                  \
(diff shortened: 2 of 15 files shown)


ginev:
https://github.com/NativeScript/ios-runtime/pull/835/commits/b62658118cff13b26dd7f57cdba694dc1adae7c0 addresses an issue with debugger delay killing the already existing listen connection at an inappropriate time. We do not need this delay to cancel the listen connection as we immediately close it after a successful frontend connection.

ginev:
@KristinaKoeva - could you please take a look at this PR and comment on the following questions:
- why do we recreate the  backend dispatchers when `didCreateFrontendAndBackend` is called and not use the pattern as seen in the WebKit's original source as pointed in Introduction section of the first comment
- regarding the handling of connection sockets - can you update us on why there is a 30s delay which closes the listening socket if no inspector is present?

KristinaKoeva:
The original webkit source was not the same at the time we wrote this logic :)
Take a look here - https://github.com/WebKit/webkit/commit/d81c28de2f870a2f3e6988d6b0d1907c1f2fb5bf

The delay is there because otherwise you will have an open socket for as long as the application lives.

Did the webkit update broke the second attach attempt ?

ginev:
@KristinaKoeva - thanks. This is helpful. The issue with the RefPtr objects comes because of the fact that we recreate the DomainBackendDispatcher each time. It registers with the main Backend Dispatcher's HashMap of dispatchers and chains its supplemental dispatchers. The first iteration of this is OK but the second one causes us to find an invalid RefPtr from the HashMap and chain it to our DomainBackendDispatcher here: https://github.com/NativeScript/ios-runtime/blob/master/src/NativeScript/inspector/DomainBackendDispatcher.cpp#L31-L33

src/debugging/TNSDebugging.h
```diff
@@ -249,39 +296,57 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
           return nil;
       }
 
-      if (inspector) {
-          return nil;
+      // Keep a working copy for calling into the VM after releasing inspectorLock
+      TNSRuntimeInspector* tempInspector = nil;
```
mbektchiev: Move declaration to assignment line.

src/debugging/TNSDebugging.h
```diff
@@ -249,39 +296,57 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
...
+          //            CFRunLoopRef runloop = CFRunLoopGetMain();
+          //            CFRunLoopPerformBlock(
+          //                                  runloop, (__bridge CFTypeRef)(NSRunLoopCommonModes), ^{
+          //                                      // If we pause right away the debugger messages that are send
+          //                                      // are not handled because the frontend is not yet initialized
+          //                                      CFRunLoopRunInMode(kCFRunLoopDefaultMode, 1, false);
+          //
+          //                                      [inspector pause];
+          //                                  });
+          //            CFRunLoopWakeUp(runloop);
```
mbektchiev: Delete commented block.

src/debugging/TNSDebugging.h
```diff
@@ -249,39 +296,57 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
...
             CFRunLoopPerformBlock(
                 runloop, (__bridge CFTypeRef)(inspectorRunloopModes), ^{
-                  [inspector dispatchMessage:message];
+                  // Keep a working copy for calling into the VM after releasing inspectorLock
+                  TNSRuntimeInspector* tempInspector = nil;
+                  @synchronized(inspectorLock()) {
+                      tempInspector = inspector;
+                  }
+
+                  [tempInspector dispatchMessage:message];
```
mbektchiev: Check whether inspector is `null`

mbektchiev:
run ci

mbektchiev:
run ci (after `git clean -fdx` in workspace)

src/NativeScript/TNSRuntime+Inspector.mm
```diff
@@ -132,7 +134,15 @@ - (void)pause {
     self->_inspectorController->pause();
 }
 
+- (bool)hasFrontends {
```
tdermendjiev: We might delete this method as it's never called.

ginev:
It might be good to rename [this](https://github.com/NativeScript/ios-runtime/blob/64a8a9bfd1938c29f0bbcbb21bf539da461fa3a7/src/debugging/TNSDebugging.h#L169) to be `receiver` as it rights to the `io` channel and is analogous to the `sender` block [here](https://github.com/NativeScript/ios-runtime/blob/64a8a9bfd1938c29f0bbcbb21bf539da461fa3a7/src/debugging/TNSDebugging.h#L133).

mbektchiev:
run ci

PR #838, opened 2017-12-14
fix(inspector): Fix API compatibility issue with Inspector

ginev:
There are internal APIs used by our WebKit build that are not present on High Sierra. We need to use the builtin WebCore frameworks to avoid crashing instead of our own. This fix checks if we are on High Sierra and resorts to the builtin frameworks if yes. We need this up until we upgrade to the latest WebKit version which does not use the internal APIs.


ginev:
run ci

abdul-elah-js:
How to update to this version 3.4.0 ?

ginev:
@abdul-elah-js - you can simply  use the `tns platorm add ios@3.4.0` command.

PR #844, opened 2017-12-20
fix(console): Fix issue with missing console.time() messages.

ginev:
This PR addresses an issue with console.time() messages missing on iOS unless the Inspector console is used. The fix consists of extending the `InspectorConsoleAgent`'s API to return the generated console.time() message which is then consumed by the `GlobalObjectConsoleClient` which prints it in the Console.

Related: https://github.com/NativeScript/ios-runtime/issues/843

### DIFF
--- a/src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
+++ b/src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
@@ -115,12 +115,20 @@ void GlobalObjectConsoleClient::takeHeapSnapshot(JSC::ExecState*, const String& 
 }
 
 void GlobalObjectConsoleClient::time(JSC::ExecState*, const String& title) {
-    m_consoleAgent->startTiming(title);
+    std::unique_ptr<String> startMsg = m_consoleAgent->startTiming(title);
+    const String* url;
+    if (startMsg) {
+        ConsoleClient::printConsoleMessage(MessageSource::ConsoleAPI, MessageType::Timing, MessageLevel::Warning, *startMsg.get(), (const String&)url, -1, -1);
+    }
 }
 
 void GlobalObjectConsoleClient::timeEnd(JSC::ExecState* exec, const String& title) {
-    RefPtr<Inspector::ScriptCallStack> callStack(Inspector::createScriptCallStackForConsole(exec, 1));
-    m_consoleAgent->stopTiming(title, WTFMove(callStack));
+    Ref<Inspector::ScriptCallStack> callStack(Inspector::createScriptCallStackForConsole(exec, 1));
+    std::unique_ptr<String> stopMsg = m_consoleAgent->stopTiming(title, WTFMove(callStack));
+    const String* url;
+    if (stopMsg) {
+        ConsoleClient::printConsoleMessage(MessageSource::ConsoleAPI, MessageType::Timing, MessageLevel::Debug, *stopMsg.get(), (const String&)url, -1, -1);
+    }
 }
 
 void GlobalObjectConsoleClient::timeStamp(JSC::ExecState*, RefPtr<Inspector::ScriptArguments>&&) {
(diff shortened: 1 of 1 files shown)


src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
```diff
@@ -115,12 +115,20 @@ void GlobalObjectConsoleClient::takeHeapSnapshot(JSC::ExecState*, const String&
 }
 
 void GlobalObjectConsoleClient::time(JSC::ExecState*, const String& title) {
-    m_consoleAgent->startTiming(title);
+    std::unique_ptr<String> startMsg = m_consoleAgent->startTiming(title);
+    const String* url;
+    if (startMsg) {
+        ConsoleClient::printConsoleMessage(MessageSource::ConsoleAPI, MessageType::Timing, MessageLevel::Warning, *startMsg.get(), (const String&)url, -1, -1);
```
Pip3r4o: According to the specification https://console.spec.whatwg.org/#time there should be no logging when `console.time()` is called, unless a timer with the same label has already been started.

ginev: @Pip3r4o - that's the case here too. The `startMsg` will always be `nullptr` unless you have called `console.time()` twice with the same token. Then a warning will be printed.

ginev:
run ci

src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
```diff
@@ -115,12 +115,20 @@ void GlobalObjectConsoleClient::takeHeapSnapshot(JSC::ExecState*, const String&
 }
 
 void GlobalObjectConsoleClient::time(JSC::ExecState*, const String& title) {
-    m_consoleAgent->startTiming(title);
+    std::unique_ptr<String> startMsg = m_consoleAgent->startTiming(title);
+    const String* url;
```
mbektchiev: `url` should be declared as `String url;` and can be moved inside the `if` block as it is not used outside. The cast to `const String &` in the call to `printConsoleMessage` should be removed as well.

src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
```diff
@@ -115,12 +115,20 @@ void GlobalObjectConsoleClient::takeHeapSnapshot(JSC::ExecState*, const String&
...
+        ConsoleClient::printConsoleMessage(MessageSource::ConsoleAPI, MessageType::Timing, MessageLevel::Warning, *startMsg.get(), (const String&)url, -1, -1);
+    }
 }
 
 void GlobalObjectConsoleClient::timeEnd(JSC::ExecState* exec, const String& title) {
-    RefPtr<Inspector::ScriptCallStack> callStack(Inspector::createScriptCallStackForConsole(exec, 1));
-    m_consoleAgent->stopTiming(title, WTFMove(callStack));
+    Ref<Inspector::ScriptCallStack> callStack(Inspector::createScriptCallStackForConsole(exec, 1));
+    std::unique_ptr<String> stopMsg = m_consoleAgent->stopTiming(title, WTFMove(callStack));
+    const String* url;
```
mbektchiev: `url` should be declared as `String url;` and can be moved inside the `if` block as it is not used outside. Also the cast to `const String &` in the call to `printConsoleMessage` should be removed as well.

src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
mbektchiev: Actually you can pass `WTF::emptyString()` without having to construct a new instance.

src/NativeScript/inspector/GlobalObjectConsoleClient.cpp
mbektchiev: Same here,  you can pass `WTF::emptyString()` without having to construct a new instance.

PR #853, opened 2018-01-05
chore: add waffle.io badge

radeva:
Add waffle.io badge to waffle board in order to give more visibility to the current issues status.


README.md
```diff
@@ -1,5 +1,7 @@
 # NativeScript for iOS Runtime
 
+[![Waffle.io - NativeScript Runtimes and CLI](https://badge.waffle.io/NativeScript/android-runtime.svg?columns=In%20Progress)](https://waffle.io/NativeScript/android-runtime)
```
mbektchiev: Typo: The url is to `android-runtime` instead of `ios-runtime`

mbektchiev: As I understand the URL is `android-runtime` but the waffle view is actually aggregating several repositories and this is the intended URL.

PR #869, opened 2018-01-25
Default value for MACOSX_DEPLOYMENT_TARGET

ginev:
This is the first step towards implementing build infrastructure that supports both Sierra and Xcode 8, and High Sierra with Xcode 9.+ for the Inspector builds. This change allows us to pass the MACOSX_DEPLOYMENT_TARGET as an external parameter from Jenkins and execute separate builds producing different artefacts.


ginev:
@Natalia-Hristova

PR #874, opened 2018-01-26
chore: Update package script to copy already built binaries.

ginev:
The $BINARIES_DIR will be defined by the build itself.


ginev:
@Natalia-Hristova

PR #876, opened 2018-02-08
Fix issues with using private Apple APIs

ginev:
After upgrading WebKit to v11.0 we've encountered several issues with using private APIs forbidden for usage by apps on the AppStore. This PR addresses these issues in the following way:

- mock the unorm2_ APIs and replace them with our own `String.prototype.normalize` implementation using the official CFString APIs. This is done in `GlobalObject.mm` whereas the mocking is done in the WebKit's source code
- replace the usage of the private Crypto APIs with public APIs

A unit test is added that verifies the behavior of `String.prototype.normalize` against the https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/normalize specification.

### DIFF
--- a/tests/TestRunner/app/FunctionsTests.js
+++ b/tests/TestRunner/app/FunctionsTests.js
@@ -18,4 +18,44 @@ describe(module.id, function () {
         expect(CFBagGetCountOfValue(bag, a)).toBe(1);
         expect(CFBagContainsValue(bag, a)).toBe(true);
     });
+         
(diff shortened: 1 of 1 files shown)


ginev:
run ci

tdermendjiev:
run ci

tdermendjiev:
run ci

Natalia-Hristova:
run ci

Natalia-Hristova:
run ci

PR #889, opened 2018-03-01
feature: SIMD matrices support

tdermendjiev:
Current implementation of libffi doesn't handle ext_vector values properly on Aarch64. New ffi type = FFI_TYPE_EXT_VECTOR is introduced in order for **ffi_call** function to be able to distinguish between standard collection types (marked as FFI_TYPE_STRUCT) and ext_vector ones as ext_vector values are not being stored in general-purpose registers.


Natalia-Hristova:
run ci

ginev:
run ci

ginev:
It seems the build fails on Xcode 8 machines.

src/NativeScript/Marshalling/Reference/ExtVectorTypeInstance.cpp
```diff
@@ -0,0 +1,121 @@
...
+}
+
+const char* ExtVectorTypeInstance::encode(VM& vm, JSCell* cell) {
+    ExtVectorTypeInstance* self = jsCast<ExtVectorTypeInstance*>(cell);
+
+    if (!self->_compilerEncoding.empty()) {
+        return self->_compilerEncoding.c_str();
+    }
+
+    self->_compilerEncoding = "^";
```
mbektchiev: According to the [Type Encodings documentation article](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/ObjCRuntimeGuide/Articles/ocrtTypeEncodings.html)
>The type code for an array is enclosed within square brackets; the number of elements in the array is specified immediately after the open bracket, before the array type. For example, an array of 12 pointers to floats would be encoded as:
[12^f]

I think that `IndexedRefTypeInstance`'s and `ExtVectorTypeInstance`'s compiler encodings should both be changed accordingly.

Natalia-Hristova:
run ci

Natalia-Hristova:
run ci

Natalia-Hristova:
run ci

PR #894, opened 2018-03-20
fix(module-loader): Fix issue with loading empty files.

ginev:


ginev:
Fixes: https://github.com/NativeScript/ios-runtime/issues/895

PR #896, opened 2018-03-23
fix: Incorrect constant array type parsing

tdermendjiev:
When arrays are passed as parameters in functions or returned by them, they are passed by reference so they should be parsed as ReferenceTypeInstance. When they are elements of other arrays/vectors or struct fields they are passed by value. In such cases they should be parsed as IndexedRefTypeInstance.


src/NativeScript/Marshalling/Reference/ExtVectorTypeInstance.cpp
```diff
@@ -91,8 +91,8 @@ void ExtVectorTypeInstance::finishCreation(JSC::VM& vm, JSCell* innerType) {
 
     ffi_type* type = new ffi_type({ .size = arraySize * innerFFIType->size, .alignment = innerFFIType->alignment, .type = FFI_TYPE_EXT_VECTOR });
 
-    ffi_type* ffiTypeEl = new ffi_type({ .size = 4,
-                                         .alignment = 4,
+    ffi_type* ffiTypeEl = new ffi_type({ .size = sizeof(float),
```
mbektchiev: `ffiTypeEl` is redundant (even wrong in case the actual inner type is `double`). The element type should be `innerFFIType` instead.

PR #931, opened 2018-05-30
feat: Add SCNVector3 and SCNVector4 inline functions

tdermendjiev:


tdermendjiev:
Inline functions will be added by plugin/use case.

PR #951, opened 2018-07-02
fix: Catch and report unhandled exceptions after ffiClosureCallback ** skip ci **

mbektchiev:
During the latest WebKit update we discovered that ffiClosureCallback
was crashing the application whenever a JS exception escaped it and
removed this behavior as undesirable.

However when the callback is made from the topmost JS frame we need
to crash since there's noone left to handle the exception. In order
to detect this situation we've enabled EXCEPTION_SCOPE_VERIFICATION
in JavaScriptCore in release builds (it was enabled only in debug by
default).

Now whenever we catch an exception we're checking whether we're at the top
before deciding whether to crash or leave it for handling in JS. The scope
is considered to be top level when the exception scope's recursion count
is either 0 or equal to the one set in UIApplicationMain(); or when
UIApplicationMain() is at the top of the callframe stack.


tdermendjiev:
Fixes #971

PR #964, opened 2018-07-31
fix(runtime): Throw an Objective C exception on fatal error

mbektchiev:
Raise an Objective C exception with JS callstack information, instead of causing
an EXC_BAD_ACCESS signal when a fatal error is detected. This way of crashing
is more appropriate for handling by error reporting analytics libraries like Sentry

refs #963

## PR Checklist

### DIFF
--- a/examples/Gameraww/app/MasterViewController.js
+++ b/examples/Gameraww/app/MasterViewController.js
@@ -21,6 +21,7 @@ var JSMasterViewController = UITableViewController.extend(
         this.loadData();
       },
       "aboutPressed:" : function(sender) {
+        throw new Error("crash!");
         var alertWindow = new UIAlertView({
           title : "About",
           message : "NativeScript Team",
--- a/src/NativeScript/JSErrors.mm
+++ b/src/NativeScript/JSErrors.mm
@@ -64,20 +64,21 @@ void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception, 
         NSLog(@"Native stack trace:");
         WTFReportBacktrace();
 
-        NSLog(@"JavaScript stack trace:");
+        std::stringstream jsCallstack;
         RefPtr<Inspector::ScriptCallStack> callStack = Inspector::createScriptCallStackFromException(execState, exception, Inspector::ScriptCallStack::maxCallStackSizeToCapture);
         for (size_t i = 0; i < callStack->size(); ++i) {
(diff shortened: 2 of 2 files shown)


NathanWalker:
You are my hero @mbektchiev 🙌

bradmartin:
Sentry now logs with: https://sentry.io/share/issue/91e007434c504ab7aef19c745ecb8fcd/ which is an improvement as this gives me something better than `EXC_BAD_ACCESS` because as a non-ios dev I didn't know what that meant, with `NSGenericException` that gives me an idea something with NativeScript. Thanks for working on this @mbektchiev - looks promising. I'll add you to the sentry test account if you'd like to view these without me sending the share link. ❤️

Natalia-Hristova:
run ci

mbektchiev:
@bradmartin Good news, I came up with another idea. I've changed the exception's name from `NSGenericException` to the whole description that I had previously put as its `reason`. I hope that this way Sentry will be able to display it. 

But nevertheless I think it's a good idea that you file an issue with them and suggest them to include the reason in their crash reports.

bradmartin:
Hey @mbektchiev - I did get a reply from someone who I think is on the Sentry team on their forums. Seemed excited about making this work 👍 (https://forum.sentry.io/t/sentry-for-ios-uncaught-exception-help-within-nativescript-framework/4367)

I had a question about the exception, in your last comment you mention `suggest them to include the reason...` and I just looked at the code https://github.com/NativeScript/ios-runtime/blob/158c95f49d447f60c012c7425a4dafd21a75b893/src/NativeScript/JSErrors.mm#L103 and this is `nil` so right now would that mean the `reason` for the NSException is empty so it wouldn't be logged? 

I just want to make sure I convey the correct terms to whoever if I need to. 

Thanks again Martin ❤️

tdermendjiev:
Fixes #969

mbektchiev:
@bradmartin Yes, you're right, `reason` is not used at the moment. When we saw that it is being dropped we decided move the call stack to the name of the exception hoping that this will make it visible in the report. As we observed, however, it was being truncated. Nevertheless we merged it this way knowing that it's not perfect. 

I think that the optimal variant would be to have the name be a one line exception string which includes the JS exception's message, and then the whole stack trace in the `reason` field. But I don't think that we should change it until Sentry starts respecting the `reason` property. And if they remove the truncation of `name` we might not even need to change it at all.

Issue #877, opened 2018-02-09
GVRKit plugin metadata generation issue


NathanaelA:
I have a GVRKit plugin that I have been developing for a project, and the runtime metadata is incorrect, and I need to find a solution to get around this in NativeScript.

In some ways slightly related to #826 (Like him I've got one that is working).
Partially related to this issue #710...

Error is:
`Error: Actual arguments count: "0". Expected: "1".`

```
      - Name:            handleTrigger
        JsName:          handleTrigger
        Filename:        GVRRenderer.h
        Module:          
          FullName:        GVRKit.GVRRenderer
          IsPartOfFramework: true
          IsSystemModule:  false
          Libraries:       
        Flags:           [ IsIosAppExtensionAvailable ]
        Type:            Method
        Signature:       
          - Type:            Bool

      - Name:            'handleTrigger:'
        JsName:          handleTrigger
        Filename:        GVRRenderer.h
        Module:          
          FullName:        GVRKit.GVRRenderer
          IsPartOfFramework: true
          IsSystemModule:  false
          Libraries:       
        Flags:           [ IsIosAppExtensionAvailable ]
        Type:            Method
        Signature:       
          - Type:            Bool
          - Type:            Interface
            JsName:          GVRHeadPose
            WithProtocols:   
```

Do you see that the metadata JS name is the SAME, but the core name is different.   One has a prototype of:
`bool handleTrigger()`
and the other  is
`bool handleTrigger(GVRHeadPose *headpose)`

I really want to call the first option.  I can't seem to because NativeScript has decided the only valid one is the second one.  ;-(

Solutions?

ulvesked:
Hi, I have the same issues with StarIO-framework.

      - Name:            'appendLineFeed:'
        JsName:          appendLineFeed
        Filename:        /Users/anders/Apps/nativescript-star-io/demo/node_modules/nativescript-star-io/platforms/ios/StarIO_Extension.framework/Headers/ISCBBuilder.h
        Module:          
          FullName:        StarIO_Extension.ISCBBuilder
          IsPartOfFramework: true
          IsSystemModule:  false
          Libraries:       
        Flags:           [ IsIosAppExtensionAvailable ]
        Type:            Method
        Signature:       
          - Type:            Void
          - Type:            Int
      - Name:            appendLineFeed
        JsName:          appendLineFeed
[...]

PR #1002, opened 2018-10-10
test: Add test for SCNVector3 parameters

tdermendjiev:
## PR Checklist

- [x] The PR title follows our guidelines: https://github.com/NativeScript/NativeScript/blob/master/CONTRIBUTING.md#commit-messages.
- [x] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.
- [x] You have signed the [CLA](http://www.nativescript.org/cla).
- [x] All existing tests are passing: https://github.com/NativeScript/ios-runtime/blob/master/CONTRIBUTING.md#contribute-to-the-code-base
- [x] Tests for the changes are included.

## What is the current behavior
While working with SceneKit inline functions it appeared that **SCNVector3ToFloat3** won't work as its **SCNVector3** parameter is not properly marshaled.

## What is the new behavior?
No new behavior. Just a test for the above mentioned scenario is added.

Related to #979

PR #1013, opened 2018-10-31
feat: Add support for methods with same name and different parameters

tdermendjiev:
A single FFICall (JSFunction) now holds a reference to a bucket of functions instead of the 1:1 FFICall - ObjCMethod we used to have. Depending on the parameters passed to the call we now decide runtime which method to invoke.|

A question worth asking is won't it be better if we swap the names of FFICall and FunctionWrapper because it would make more sense for the FunctionWrapper to have a vector of FFICalls (each FFICall with it's cif) instead of the contrary.

## PR Checklist

In Objective-C two methods with the same name but with different set of parameters would have different selectors (e.g. "baseMethod" and "baseMethod:") but they will still have the same **jsName** (which we call from js - e.g. "baseMethod" and "baseMethod"). Because of the 1:1 mapping of the FFICall and the ObjC method we are able to call only one of the methods. 

## What is the new behavior?
A check for the parameters is added so we know exactly which method to call runtime.

Implements #877 .


tests/TestRunner/app/MethodCallsTests.js
```diff
@@ -181,6 +200,13 @@ describe(module.id, function () {
         var actual = TNSGetOutput();
         expect(actual).toBe("instance baseMethod called");
     });
+    it('Derived_OverloadedInstanceBaseMethod', function () {
+        var instance = TNSDerivedInterface.alloc().init();
+        instance.baseMethod(1);
+       
+        var actual = TNSGetOutput();
+        expect(actual).toBe("overloaded instance baseMethod: called");
```
mbektchiev: I don't see this method being overridden in `TNSDerivedInterface`, and the string that you are expecting is actually the one from `TNSBaseInterface `

tests/TestRunner/app/index.js
```diff
@@ -1,3 +1,6 @@
+
+
+
 // Inform the test results runner that the runtime is up.
```
mbektchiev: Forgotten empty lines, you can revert them.

Natalia-Hristova:
test

Natalia-Hristova:
test --performance

src/NativeScript/ObjC/Constructor/ObjCConstructorNative.mm
```diff
@@ -38,13 +38,25 @@
...
-        SymbolLoader::instance().ensureModule(method->topLevelModule());
+    std::vector<const MethodMeta*> methods = constructor->_metadata->getStaticMethods(propertyName.publicName());
+
+    if (methods.size() > 0) {
+
+        std::unordered_map<std::string, std::vector<const MethodMeta*>> metasByJsName = Metadata::getMetasByJSNames(methods);
+
+        for (auto& methodName : metasByJsName) {
+
+            for (const MethodMeta* memberMeta : metasByJsName[methodName.first]) {
```
mbektchiev: This loop is unneeded as well.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -112,6 +121,48 @@ static int compareIdentifiers(const char* nullTerminated, const char* notNullTer
...
+            }
+        }
+    }
+
+    // search in protcols
+    if (includeProtocols) {
+        for (Array<String>::iterator it = protocols->begin(); it != protocols->end(); ++it) {
+            const ProtocolMeta* protocolMeta = reinterpret_cast<const ProtocolMeta*>(MetaFile::instance()->globalTable()->findMeta((*it).valuePtr()));
+            if (protocolMeta != nullptr) {
+                if (const MemberMeta* method = protocolMeta->member(identifier, length, type, onlyIfAvailable)) {
```
mbektchiev: Recursively call `members` in order to retrieve all matching protocol members.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -112,6 +121,48 @@ static int compareIdentifiers(const char* nullTerminated, const char* notNullTer
...
+        break;
+    default:
+        members = nullptr;
+        ASSERT_NOT_REACHED();
+    }
+
+    std::vector<const MethodMeta*> result;
+    if (members != nullptr) {
+
+        for (int i = 0; i < members->count; i++) {
```
mbektchiev: Use `binarySearch` for finding a member with the requested name and then linearly include all others surrounding it. You can implement this logic in a new `binarySearchRange` function in `Array`.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -112,6 +121,48 @@ static int compareIdentifiers(const char* nullTerminated, const char* notNullTer
...
+
+            const MethodMeta* m = reinterpret_cast<const MethodMeta*>((*members)[i].valuePtr());
+            if ((strncmp(identifier, m->jsName(), length) == 0) && strlen(m->jsName()) == length) {
+                result.push_back(m);
+            }
+        }
+    }
+
```
mbektchiev: I think we may need to also scan the members of base classes. But I'm not sure if they're present in the current class' members or not. You should write a unit test and make it pass if it doesn't.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -111,6 +118,63 @@ static int compareIdentifiers(const char* nullTerminated, const char* notNullTer
 
     return nullptr;
 }
+const MethodMeta* BaseClassMeta::member(const char* identifier, size_t length, MemberType type, size_t paramsCount, bool includeProtocols, bool onlyIfAvailable) const {
+    const std::vector<const MethodMeta*> metas = this->members(identifier, length, type, includeProtocols, onlyIfAvailable);
```
mbektchiev: `const MemberMeta* BaseClassMeta::member(const char* identifier, size_t length, MemberType type, bool includeProtocols, bool onlyIfAvailable) const {` can be implemented by calling `this->members` as well. And by doing so we will avoid a lot of duplicated code.
I know that if we call directly the current implementation it may have some performance hit, but let's measure the impact of the whole implementation and then decide what and where to optimize.

src/NativeScript/Metadata/Metadata.mm
mbektchiev: Inherited class members are handled via the prototype chain and need not be considered here. Resolving comment.

src/NativeScript/Metadata/Metadata.h
```diff
@@ -18,6 +18,21 @@ namespace Metadata {
...
 
+template <typename V>
+static const V& getProperFunctionFromContainer(const std::vector<V>& container, int argsCount, std::function<int(const V&)> paramsCounter) {
+    const V* callee = nullptr;
+
+    for (const V& func : container) {
+        if (paramsCounter(func) == argsCount) {
+            callee = &func;
+            break;
+        } else if (!callee || (callee && paramsCounter(*callee) > paramsCounter(func) && paramsCounter(func) > argsCount)) {
```
mbektchiev:
```
auto candidateArgs = paramsCounter(func);
auto calleeArgs = paramsCounter(*callee);

if (candidateArgs == argsCount) {
   callee = &func;
   break;
} else if (!callee) {
  // no candidates so far, take it whatever it is
  callee = &func;
} else if (argsCount < candidateArgs && (calleeArgs < argsCount || candidateArgs < calleeArgs)) {
  // better candidate - looking for the least number of arguments which is more than the amount actually passed
  callee = &func;
} else if (calleeArgs < candidateArgs) {
  // better candidate - looking for the maximum number of arguments which less than the amount actually passed (if one with more cannot be found)
  callee = &func;
}
```

tdermendjiev:
test

mbektchiev:
Merged with #1002  and #1024

tdermendjiev:
test

tdermendjiev:
test

tdermendjiev:
test

PR #1034, opened 2018-12-13
Tdermendzhiev/wip f overloading protocol methods issue

tdermendjiev:
A separate PR for various fixes:
*ObjcExtends refactoring
*fix calling base class methods
*tests for *ObjCPrototype.put* and explicitly calling methods by *BaseClass.call(instance)*

### DIFF
--- a/src/NativeScript/Metadata/Metadata.mm
+++ b/src/NativeScript/Metadata/Metadata.mm
@@ -96,35 +96,71 @@ const MethodMeta* BaseClassMeta::member(const char* identifier, size_t length, M
     return (MethodMeta*)result;
 }
 
-const std::vector<const MemberMeta*> BaseClassMeta::members(const char* identifier, size_t length, MemberType type, bool includeProtocols, bool onlyIfAvailable) const {
+void collectInheritanceChainMembers(const char* identifier, size_t length, MemberType type, bool onlyIfAvailable, const BaseClassMeta* derivedClass, std::function<void(const MemberMeta*)> collectMember) {
 
     const ArrayOfPtrTo<MemberMeta>* members = nullptr;
     switch (type) {
     case MemberType::InstanceMethod:
-        members = &this->instanceMethods->castTo<PtrTo<MemberMeta>>();
+        members = &derivedClass->instanceMethods->castTo<PtrTo<MemberMeta>>();
         break;
     case MemberType::StaticMethod:
-        members = &this->staticMethods->castTo<PtrTo<MemberMeta>>();
+        members = &derivedClass->staticMethods->castTo<PtrTo<MemberMeta>>();
         break;
     case MemberType::InstanceProperty:
-        members = &this->instanceProps->castTo<PtrTo<MemberMeta>>();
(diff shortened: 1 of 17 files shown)


src/NativeScript/ObjC/ObjCMethodCallback.h
```diff
@@ -49,6 +48,42 @@ class ObjCMethodCallback : public FFICallback<ObjCMethodCallback> {
 
     bool _hasErrorOutParameter;
 };
+
+inline void overrideObjcMethodCalls(ExecState* execState, JSObject* object, PropertyName propertyName, JSCell* method, const Metadata::BaseClassMeta* meta, Metadata::MemberType memberType, Class klass,
```
mbektchiev: We'd better move the implementation of this method out of the header file. The reason I initially put it here was that I expected it to be a template method.

tests/TestRunner/app/ApiTests.js
```diff
@@ -201,7 +201,7 @@ describe(module.id, function () {
         expect(TNSApi.alloc().init().strokeColor).toBeNull();
     });
 
-    if (TNSIsConfigurationDebug) {
+    if (false && TNSIsConfigurationDebug) {
```
mbektchiev: The `false &&` part was to disable these lengthy tests while developing, we should remove it.

tests/TestRunner/app/ApiTests.js
```diff
@@ -556,7 +558,7 @@ describe(module.id, function () {
         expect(result).toBe('method called');
     });
 
-    if (TNSIsConfigurationDebug) {
+    if (false && TNSIsConfigurationDebug) {
```
mbektchiev: The false && part was to disable these lengthy tests while developing, we should remove it.

src/NativeScript/Metadata/Metadata.mm
```diff
@@ -96,6 +96,35 @@ static int compareIdentifiers(const char* nullTerminated, const char* notNullTer
     return (MethodMeta*)result;
 }
 
+const std::vector<const MemberMeta*> BaseClassMeta::collectInheritanceChainMembers(const char* identifier, size_t length, MemberType type, bool includeProtocols, bool onlyIfAvailable, const ArrayOfPtrTo<MemberMeta>* members, int resultIndex) const {
```
mbektchiev: There's too much code duplication between this method and `members`. Try to eliminate it by implementing all cases of `members` in terms of the new `collectInheritanceChainMembers`.

src/NativeScript/ObjC/Inheritance/ObjCClassBuilder.h
```diff
@@ -46,6 +47,9 @@ class ObjCClassBuilder {
 
     ObjCConstructorDerived* build(JSC::ExecState*);
 
+private:
+    void addMethod(JSC::ExecState* execState, ObjCMethodWrapper* wrapper, const JSC::Identifier& jsName, JSC::JSCell* method);
```
mbektchiev: This method seems to be unused, we should remove it if that's true.

tdermendjiev:
test

Issue #620, opened 2016-08-10
[Proposal]: WebWorkers support


ivanbuhov:
#### Description
General guidelines for the Workers implementation effort and specification in the context of NativeScript. 
We have an existing issue [here](https://github.com/NativeScript/NativeScript/issues/85), but the purpose here is to show the road map that the team intends to follow in developing the Workers functionality. The general scenarios we want to support are listed at the bottom.

#### Limitations 
In NativeScript we don’t need to implement all details of the [web workers specification](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers), because some of these details are only related to the browser, and won’t have any meaning in the context of a NativeScript application. The features and syntax that will follow has the purpose of describing the adoption of the web workers specification in NativeScript. In this document we will describe and list what we intend to support.

#### Guidelines from the specification to follow

_context_
>Notice that onmessage and postMessage() need to be hung off the Worker object when used in the main script thread, but not when used in the worker. This is because, inside the worker, the worker is effectively the global scope.

_messaging specifics_
>Data passed between the main thread and workers is copied, not shared. Objects are serialized as they're handed to the worker, and subsequently, deserialized on the other end. The page and worker do not share the same instance, so the end result is that a duplicate is created on each end. Most browsers implement this feature as structured cloning.
The [structured cloning](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Structured_clone_algorithm) algorithm can accept JSON and a few things that JSON can't — like circular references.

_messaging transfer ownership_
>Passing data by transferring ownership (transferable objects): in nativescript we can try to pass native objects by transferring ownership on them.

_onerror spec_
>The error event has the following three fields that are of interest:
message: A human-readable error message.
filename: The name of the script file in which the error occurred.
lineno: The line number of the script file on which the error occurred.

_importScripts_
>_ImportScripts_ in a nativescript context mean _require_. If we want to support this syntax we need an alias for require.

#### Syntax

The following example will be the general syntax to use when working with Workers. Syntax is based on [web worker specification](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Using_web_workers).

_<app_name>/app/main.js_
```javascript
// create new worker
var myWorker = new Worker("worker.js");

// on worker error handler
myWorker.onerror = function (e) {
 //do on worker error
}

// receive messages from worker
myWorker.onmessage = function (e) {
  result.textContent = e.data;
}
....
[...]

Issue #921, opened 2018-04-26
Interop.Pointer int64 creation from NativeScript side


NathanaelA:
### Please, provide the following version numbers that your issue occurs with:
3.4.x  (Probably 4.x also)

### Is there any code involved? 
var x = new Interop.Pointer(-1) =  `<Pointer: 0xffffffff>`
However, returning a pointer from ObjC of -1 = `<Pointer: 0xffffffffffffffff>`

Is their some secret to create a Interop.Pointer on the Javascript side that is the full int64 size, as it is a special value in a plugin I'm working on.   This is the only piece that I'm having to create a tiny OBJC wrapper for; and I would like to eliminate this wrapper if possible.

charsleysa:
@NathanaelA looks like they are using 32bit transformations behind the scenes.
https://github.com/NativeScript/ios-runtime/blob/9d967a09370bee938b5f8a16f9b41bbc696555ab/src/NativeScript/Marshalling/Pointer/PointerPrototype.cpp

At this point I don't think you'll be able to create 64bit pointers from JS.

NathanaelA:
@charsleysa - Stefan, yeah that was my determination also.    Hence this issue to hopefully have them create a new `Interop.Pointer64` method to do so. :grinning:

mbektchiev:
I tested and it seems that now you can create a pointer from a 64bit number. 
```JS
console.log(new interop.Pointer(-2));
console.log(new interop.Pointer(-1));
```
Outputs:
```
CONSOLE LOG file:///app/index.js:5:12: <Pointer: 0xfffffffffffffffe>
CONSOLE LOG file:///app/index.js:6:12: <Pointer: 0xffffffffffffffff>
```

I found another issue, however -- `-1` turned out to be a special value for deleted keys in the [JSC::WeakGCMap](https://github.com/NativeScript/ios-runtime/blob/master/src/NativeScript/Interop.h#L90) that we're using and causes an assertion failure in debug builds of the runtime. Meaning that it still is not completely safe, regardless that it appears to work. I plan to fix it this week.

PR #1040, opened 2018-12-20
fix(interop): Correctly initialize pointers from numbers on 64-bit archs

mbektchiev:
JavaScript represents 64-bit integers as floating point `double` values.
The conversion from JSValue to pointer must be done by
extracting the `double` value and then converting it to integer.

**Caution:** This means that pointers with more than 54 significant bits
are likely to be rounded and misrepresented!
However, current OS and hardware implementations are using 48 bits,
so we're safe at the time being.
See https://en.wikipedia.org/wiki/X86-64#Virtual_address_space_details
and https://en.wikipedia.org/wiki/ARM_architecture#ARMv8-A

## PR Checklist

The bug has been discovered while looking for a way to pass pointers to Objective-C objects to workers. 
See https://github.com/NativeScript/ios-runtime/issues/620#issuecomment-448200182


tdermendjiev (approved):

NathanaelA:
@mbektchiev is there a way to force the pointer to be created as 64bit.     I basically need a 64 bit pointer   #921 of the value (-1).      

   Brainstrorming; could I create a 33 bit number, then subtract it + 1?

Issue #1029, opened 2018-12-05
iOS exceptions are not getting caught by uncaughtErrorEvent


tsonevn:
@KkevinLi commented on [Wed Dec 05 2018](https://github.com/NativeScript/NativeScript/issues/6662)

**Environment**
```⚠ Update available for component nativescript. Your current version is 5.0.1 and the latest available version is 5.0.3.
✔ Component tns-core-modules has 5.0.5 version and is up to date.
✔ Component tns-android has 5.0.0 version and is up to date.
✔ Component tns-ios has 5.0.0 version and is up to date.
```

**Describe the bug**

No uncaughtError event is raised for native iOS errors such as NSInvalidArgumentException.

**To Reproduce**

I am using the following to test:
```  // hard crash -[NSArray objectAtIndex:]: method sent to an uninitialized immutable array object
  static iosCrashTest1() {
    console.log("IOS Crash Test 1");
    const foo: NSArray<any> = NSArray.alloc();
    foo.objectAtIndex(3);
  }
```

**Expected behavior**

Expected an uncaughtErrorEvent to be thrown (logging purposes). Right now I get a Javascript call stack followed by *** Terminating app due to uncaught exception 'NSInvalidArgumentException' ...***

---

@tsonevn commented on [Wed Dec 05 2018](https://github.com/NativeScript/NativeScript/issues/6662#issuecomment-444416147)

Hi @KkevinLi,
We have checked your case and found that the iOS exceptions are not getting caught by `uncaughtErrorEvent`, because of a missing feature. Regarding that, I will mark the issue as a feature request and will move it in the ios-runtime repository, where we can keep the discussion about this feature. 
Also, there is an experimental branch, which enables this functionality in NativeScript. You can find it [here](https://github.com/NativeScript/ios-runtime/tree/bektchiev/throw-objc-exceptions-to-js).

mbektchiev:
Hi @KkevinLi, thank you for raising this issue. We've also thought about rethrowing Objective-C exceptions to JavaScript before. In fact there's [a branch](https://github.com/NativeScript/ios-runtime/tree/bektchiev/throw-objc-exceptions-to-js) that experiments with it. We may include it in some of the next versions if there's demand for it but the reasons we decided to not merge the branch was that generally Apple advise developers to not continue execution after catching an Objective-C exception.

KkevinLi:
Thank you, I will check out that branch.

[...]

PR #1043, opened 2018-12-21
fix(bridge): Catch ObjC Exceptions and throw them to JS

mbektchiev:
* Wrap `ffi_call`s with @try-block
* Throw ObjC exception to JavaScript
(attaching the original native one as `nativeException` property)
* Update libffi with the latest master in an effort to get the
fix from https://github.com/libffi/libffi/issues/418

refs #1029

## PR Checklist

Objective-C exceptions in calls made from JS crash the application.

## What is the new behavior?
Objective-C exceptions are intercepted by iOS Runtime and are thrown to JS, where they can be caught. If no catch handler is found they will be treated as unhandled JavaScript exceptions.

### DIFF
--- a/src/NativeScript/CMakeLists.txt
+++ b/src/NativeScript/CMakeLists.txt
@@ -118,7 +118,7 @@ set(SOURCE_FILES
     Calling/FFICallPrototype.cpp
     Calling/CFunctionWrapper.mm
     Calling/FFIFunctionCallback.cpp
-    Calling/FunctionWrapper.cpp
+    Calling/FunctionWrapper.mm
     GlobalObject.mm
     GlobalObject.moduleLoader.mm
     inspector/CachedResource.mm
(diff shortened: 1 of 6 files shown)


farfromrefug:
@mbektchiev I can't get this to work with my native call.
I am using ```ios-runtime``` version ```6.0.0-2019-05-16-165318-01``` 
I do something like this:
```typescript
try {
                const result = NTMBVectorTileDecoder.alloc().initWithCompiledStyleSet(vectorTileStyleSet);
                return result;
            } catch (err) {
                console.error('MBVectorTileDecoder error', err);
                return null;
            }
```

But it still crashes the app when the NSException is thrown
Screen Shot 2019-05-17 at 12 13 35
Any idea?

mbektchiev:
This happens only on Simulator due to this bug: https://github.com/NativeScript/ios-runtime/issues/1044

farfromrefug:
@mbektchiev indeed ! thanks

Issue #1051, opened 2019-01-08
Add OnDiscardedError handler


vtrifonov:
**Is your feature request related to a problem? Please describe.**
Currently we have a custom flag which can be used to avoid application crashes when JS exceptions are thrown [discardUncaughtJsExceptions](https://docs.nativescript.org/core-concepts/android-runtime/advanced-topics/custom-flags#discarding-javascript-exceptions-when-called-from-native). Unfortunately those type of exceptions cannot be collected through analytics, so it would be good if we can provide a way of handling such exceptions so that you could be able to get more details about where and how often they happen. 

**Describe the solution you'd like**
Adding **__onDiscardedError** similar to **__onUncaughtError** handler which will receive those exceptions can be useful to process them, e.g.:
```JavaScript
global.__onDiscardedError = function(error){
    console.log(error.message);
    console.log(error.stackTrace);
    console.log(error.nativeException);
    //report the exception in your analytics solution here
}
```

PR #1052, opened 2019-01-08
feat: OnDiscardedError handler

vtrifonov:
Related to #1051

### DIFF
--- a/src/NativeScript/JSErrors.mm
+++ b/src/NativeScript/JSErrors.mm
@@ -49,14 +49,25 @@ void reportErrorIfAny(JSC::ExecState* execState, JSC::CatchScope& scope) {
             id discardExceptionsValue = [[TNSRuntime current] appPackageJson][@"discardUncaughtJsExceptions"];
             bool discardExceptions = [discardExceptionsValue boolValue];
             scope.clearException();
-            reportFatalErrorBeforeShutdown(execState, exception, true, discardExceptions);
+            if (discardExceptions) {
+                reportDiscardedError(execState, exception);
+            } else {
+                reportFatalErrorBeforeShutdown(execState, exception, true);
+            }
         }
     }
 }
-
-void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception, bool callUncaughtErrorCallbacks, bool dontCrash) {
+    
+void reportDiscardedError(ExecState* execState, Exception* exception) {
+    NakedPtr<Exception> errorCallbackException;
     GlobalObject* globalObject = static_cast<GlobalObject*>(execState->lexicalGlobalObject());
+    globalObject->callJsDiscardedErrorCallback(execState, exception, errorCallbackException);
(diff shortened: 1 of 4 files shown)


src/NativeScript/JSErrors.mm
```diff
@@ -35,8 +35,55 @@ void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
 using namespace JSC;
 using namespace Inspector;
 
+Boolean setStackTraceProperty(JSC::ExecState* execState, JSC::Exception* exception, std::string stackTraceContent) {
```
mbektchiev: You'd better use `bool` instead of `Boolean` - it's the type `putDirect` returns.

src/NativeScript/JSErrors.mm
```diff
@@ -35,8 +35,55 @@ void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
...
+    auto error = exception->value().toObject(execState);
+    
+    Identifier stackTraceName = Identifier::fromString(execState, "stackTrace");
+    VM* vm = &execState->vm();
+    JSString* stackTrace = jsString(vm, stackTraceContent.c_str());
+    PutPropertySlot slot(error);
```
mbektchiev: `slot` is unused -> you can delete this line

src/NativeScript/JSErrors.mm
```diff
@@ -35,8 +35,55 @@ void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
...
+    
+    auto result = error->putDirect(*vm, stackTraceName, JSValue(stackTrace));
+    
+    return result;
+}
+
+std::string getNativeStackTrace(JSC::ExecState* execState, JSC::Exception* exception) {
+    Ref<Inspector::ScriptCallStack> nativeCallStack = ScriptCallStack::create();
+
```
mbektchiev: Add a comment stating where this code comes from. E.g. `// code taken from WTFReportBacktrace function`

src/NativeScript/JSErrors.mm
```diff
@@ -49,12 +96,23 @@ void reportErrorIfAny(JSC::ExecState* execState, JSC::CatchScope& scope) {
...
+            }
         }
     }
 }
 
-void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception, bool callUncaughtErrorCallbacks, bool dontCrash) {
+void reportDiscardedError(ExecState* execState, Exception* exception) {
+    NakedPtr<Exception> errorCallbackException;
+    GlobalObject* globalObject = static_cast<GlobalObject*>(execState->lexicalGlobalObject());
+    globalObject->callJsDiscardedErrorCallback(execState, exception, errorCallbackException);
```
mbektchiev: We should check whether `errorCallbackException` is set and log it. This way if the handler throws an error for some reason users will be notified. Right now they cannot detect it. We should do the same in `reportFatalErrorBeforeShutdown` as well.

PR #1128, opened 2019-04-17
Support Swift files in the .xcodeproj

tdermendjiev:
*Redirect LD to nsld.sh so we can hook between compiling and linking and generate metadata there (we need appname-Swift.h to be created (and a modulemap for it) so the metadata-generator finds it)
*Set Swift language version
*Set Strip swift symbols to NO
*Add ObjC bridging header so the Swift classes are added to the generated   appname-Swift.h

## PR Checklist

Swift files cannot be added to an .xcodeproj and accessed by the NativeScript runtime for several reasons:
- Swift symbols are stripped
- Swift language version is undefined in the project settings
- No metadata is generated for the Swift symbols accessible from the ObjC runtime

## What is the new behavior?
After all source files are compiled nsld.sh creates a modulemap pointing to the appname-Swift.h where the generated interface for the Swift classes is. As the modulemap's directory is added in the HEADER_SEARCH_PATHS the metadata-generator (called from nsld.sh) will find the header and generate metadata for it. Finally the linking is executed.

This way all Swift symbols accessible from the ObjC runtime are now accessible from NativeScript.

Implements: https://github.com/NativeScript/nativescript-cli/issues/4503


build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -278,9 +280,14 @@
 				GCC_WARN_UNINITIALIZED_AUTOS = YES;
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
+				HEADER_SEARCH_PATHS = "";
```
mbektchiev: This line should be absent.

build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -312,8 +319,13 @@
...
```
mbektchiev: This one as well.

build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -331,6 +343,10 @@
 					"$(inherited)",
 					"DEBUG=1",
 				);
+				HEADER_SEARCH_PATHS = (
```
mbektchiev: We'd better leave `"$(SRCROOT)/internal" here out of compatibility reasons. "$(SRCROOT)/internal/TKLiveSync/include"` can be left deleted.

build/project-template/__PROJECT_NAME__.xcodeproj/project.pbxproj
```diff
@@ -346,6 +362,10 @@
 				ENABLE_BITCODE = NO;
 				GCC_PRECOMPILE_PREFIX_HEADER = YES;
 				GCC_PREFIX_HEADER = "__PROJECT_NAME__/__PROJECT_NAME__-Prefix.pch";
+				HEADER_SEARCH_PATHS = (
```
mbektchiev: We'd better leave `"$(SRCROOT)/internal" here out of compatibility reasons. "$(SRCROOT)/internal/TKLiveSync/include"` can be left deleted.

build/project-template/internal/nativescript-pre-build
```diff
@@ -1,6 +1,2 @@
 #!/usr/bin/env bash
-set -e
-
-pushd "$SRCROOT/internal/metadata-generator/bin"
-./build-step-metadata-generator.py
-popd
+export >./env_vars.sh
```
mbektchiev: Which directory will be the current working dir. Shouldn't we specify a full path to some intermediates directory here?

build/project-template/internal/nsld.sh
```diff
@@ -0,0 +1,53 @@
+ #/bin/bash
+
+ set +o posix
```
mbektchiev: Maybe we should add a comment here. E.g. something like `Ignore errors while inserting environment variables (some could be readonly)`

build/project-template/internal/nsld.sh
```diff
@@ -0,0 +1,53 @@
...
+      if [ $count == 1 ]
+      then
+          HEADER_PATH=`find $DERIVED_SOURCES -name *-Swift.h`
+          DELETE_SWIFT_MODULES_DIR
+
+          mkdir -p $MODULES_DIR
+          CONTENT="module nsswiftsupport { \n header \"$HEADER_PATH\" \n export * \n}"
+          printf "$CONTENT" > "$MODULES_DIR/module.modulemap"
+      else
+      echo "-Swift.h not found!"
```
mbektchiev: This output will be visible in the whole build log, we should make it easier to understand its origins and meaning. I'd suggest something like: `NSLD: Swift bridging header "*-Swift.h" not found`

build/project-template/internal/nsld.sh
```diff
@@ -0,0 +1,53 @@
...
+      fi
+  else
+    echo "Derived sources directory not found!"
```
mbektchiev: Something like: `NSLD: Derived sources directory not found` (there's no need of `!` at the end)

build/project-template/internal/nativescript-pre-build
mbektchiev: Environment variables are not set and we need this file to revive them. That's why we can't use any of the intermediate dirs - we simply don't know them here.

Issue #1135, opened 2019-04-30
[FR] error stack : same format as android and prepareStackTrace support


farfromrefug:
I am working on  sourcemap true support for {N}.
Right you can generate sourcemap but there is no way to use them in a dev env to actually see the source of the error.

On android it is already doable using [source-map-support](https://github.com/evanw/node-source-map-support).
For it to work on iOS we would need 2 things:
- we need to have a ```prepareStackTrace``` support for the ```Error``` class like it exists in v8. It is used [here](https://github.com/evanw/node-source-map-support/blob/master/source-map-support.js#L388) in ```source-map-support``` to "translate" the stack.
- the stack format should be the same as android (also as node and browsers) with ```at ...```

PR #1180, opened 2019-08-01
feat: Error.prepareStackTrace method support

vtrifonov:
Related to #1135

Added support for Error.prepareStackTrace method similar to the one in V8 and changed the format of the stack trace to be the same as android (also as node and browsers) with ```at ...```

## PR Checklist

### DIFF
--- a/src/NativeScript/JSErrors.mm
+++ b/src/NativeScript/JSErrors.mm
@@ -23,6 +23,7 @@
 #include <JavaScriptCore/ScriptCallStack.h>
 #include <JavaScriptCore/ScriptCallStackFactory.h>
 #include <JavaScriptCore/interpreter/Interpreter.h>
+#include <JavaScriptCore/runtime/StackFrame.h>
 
 static TNSUncaughtErrorHandler uncaughtErrorHandler;
 void TNSSetUncaughtErrorHandler(TNSUncaughtErrorHandler handler) {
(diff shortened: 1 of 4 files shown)


farfromrefug:
@vtrifonov @mbektchiev First thank for that PR!
Now i think we need to discuss a bit more on how to actually use this for the user.
It is not easy to actually make it all work. But i managed to end up with this:
Screen Shot 2019-08-23 at 14 18 08
The important thing is that i get real ts file absolute path. The great thing about this is that it allow ctrl+click to go the line/column. So useful in dev!

Now there are a few steps to make it work:
* have webpack to generate source-map correctly:
    * you need to add options to ```config.output``` (it should work with source-map, inline-source-map ...)
```ts
devtoolModuleFilenameTemplate: info => {
                return info.absoluteResourcePath.split('?')[0];
            },
            devtoolFallbackModuleFilenameTemplate: info => {
                return info.absoluteResourcePath.split('?')[0];
            }
```
* have your app correctly handle source maps. I do this only in dev by stripping the code with webpack in prod
```ts
const currentApp = knownFolders.currentApp();
process.cwd = function() {
    return '';
}
require('source-map-support').install({
    environment: 'node',
    handleUncaughtExceptions: false,
    retrieveSourceMap(source) {
        const sourceMapPath = source + '.map';
        const appPath = currentApp.path;
        let sourceMapRelativePath = sourceMapPath
            // .replace('file:///', '')
            .replace('file://', '')
            .replace(appPath + '/', '')
            .replace(appPath + '/', '');
        if (sourceMapRelativePath.startsWith('app/')) {
            sourceMapRelativePath = sourceMapRelativePath.slice(4);
        }
        return {
            url: sourceMapRelativePath,
            map: currentApp.getFile(sourceMapRelativePath).readTextSync()
        };
    }
});
```
Multiple things there:
* i define ```process.cwd()``` because ```source-map-support``` requires it
* i define ```retrieveSourceMap``` to recover the file when using ```source-map```.  there i need to cleanup the path to be able to read the file.

That should be good enough. For now i got this to work perfectly on Android.

Now some issues:
* I dont seem to be able to make it work on iOS. any JS error will make the app crash with sig 11 and thus wont get in ```prepareStackTrace``` (using next runtime to have ```prepareStackTrace```)
* android: there is an issue with "native" errors where they can't be handled. not even sure they are going through prepareStackTrace as they are created on the native side. But the error that i get on the JS side has the JS stacktrace in "stacktrace" property instead of "stack".

Maybe we can discuss all that on slack. Would love to make it work consistently on iOS and Android. I could them publish a ```nativescript-source-map-support``` plugin. 

Would it be possible for wepback templates to be updated to support this?

PR #1191, opened 2019-08-22
feat: support app refresh through the notifications service

DimitarTachev:
## PR Checklist

The application refresh is supported only with the `Page.reload` debug protocol message through the debug socket.

## What is the new behavior?
The application refresh is also supported by the iOS notifications service and you are not forced to start a debug session in order to refresh the application.

## Additional context
- The code is copied from the LiveSync plugin of the Preview app and we will also reuse this notification there.

- We need this feature in order to release the Wi-Fi iOS devices story with HMR support as we will not be able to connect to the debug socket of Wi-Fi devices in the initial version.

- The initial HMR update will be faster because we will not create debug socket connections.

- The users will not see the strange `NativeScript debugger attached` message during the first HMR update.

Related to: https://github.com/NativeScript/nativescript-cli/issues/4252

### DIFF
--- a/src/debugging/TNSDebugging.h
+++ b/src/debugging/TNSDebugging.h
@@ -23,6 +23,8 @@
 // Parsing regex is 'NativeScript debugger has opened inspector socket on port (\d+) for (.*)\.'
 // It's defined in 'ios-log-parser-service.ts'
 #define LOG_DEBUGGER_PORT NSLog(@"NativeScript debugger has opened inspector socket on port %d for %@.", currentInspectorPort, [[NSBundle mainBundle] bundleIdentifier])
+#define LOG_SUCCESSFULL_REFRESH NSLog(@"Successfully refreshed the application with RefreshRequest.")
+#define LOG_FAILED_REFRESH NSLog(@"Failed to refresh the application with RefreshRequest.")
 
 // Synchronization object for serializing access to inspector variable and data socket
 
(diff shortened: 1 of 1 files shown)


src/debugging/TNSDebugging.h
```diff
@@ -428,6 +430,36 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
...
+            @try
+            {
+                JSGlobalContextRef context = runtime.globalContext;
+                JSObjectRef globalObject = JSContextGetGlobalObject(context);
+                JSStringRef liveSyncMethodName = JSStringCreateWithUTF8CString("__onLiveSync");
+                JSValueRef liveSyncMethod = JSObjectGetProperty(context, globalObject, liveSyncMethodName, NULL);
+                JSStringRelease(liveSyncMethodName);
+                if (JSValueIsUndefined(context, liveSyncMethod))
+                {
+                    LOG_FAILED_REFRESH;
```
mbektchiev: You should return after logging the failure. Also, in order to make diagnostics easier you'd better log the reason for failure as well (possibly as a separate log message so that you won't have to handle it in the CLI's parser).

src/debugging/TNSDebugging.h
```diff
@@ -428,6 +430,36 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
...
+                }
+                
+                JSValueRef result = JSObjectCallAsFunction(context, (JSObjectRef)liveSyncMethod, NULL, 0, NULL, NULL);
+                if (JSValueIsNull(context, result))
+                {
+                    LOG_FAILED_REFRESH;
```
mbektchiev: You should return after logging the failure. Also, in order to make diagnostics easier you'd better log the reason for failure as well (possibly as a separate log message so that you won't have to handle it in the CLI's parser).

src/debugging/TNSDebugging.h
```diff
@@ -428,6 +430,36 @@ static void TNSEnableRemoteInspector(int argc, char** argv,
...
+                }
+                
+                LOG_SUCCESSFULL_REFRESH;
+            }
+            @catch (NSException *exception)
+            {
+                LOG_FAILED_REFRESH;
```
mbektchiev: In order to make diagnostics easier you'd better log the reason for failure as well (possibly as a separate log message so that you won't have to handle it in the CLI's parser).

PR #1211, opened 2019-10-14
fix(runtime): Do not lose exception in TS extends

mbektchiev:
* fix(JSC): Use `SuspendExceptionScope` instead of clearing exception
* fix(runtime): Do not call initializer if `alloc` throws a JS exception
* Add tests for erroneous TS extends
* Add `tsconfig.json` and transpile `TypeScriptTests.ts`

## PR Checklist

- [ ] There is an issue for the bug/feature this PR is for. To avoid wasting your time, it's best to open a suggestion issue first and wait for approval before working on it.

Exceptions raised during TS extends call are lost and errors are difficult to detect.

## What is the new behavior?
Exceptions raised during TS extends call are correctly thrown in JS and can be processed in a catch block or unhandled exception handler, or crash with a meaningful error.

### DIFF
--- a/tests/TestRunner/app/Inheritance/TypeScriptTests.js
+++ b/tests/TestRunner/app/Inheritance/TypeScriptTests.js
@@ -1,4 +1,30 @@
-var TSObject = (function (_super) {
+// TODO: Use TypeScript definitions when they get ready
+var __extends = (this && this.__extends) || (function () {
+    var extendStatics = function (d, b) {
+        extendStatics = Object.setPrototypeOf ||
+            ({ __proto__: [] } instanceof Array && function (d, b) { d.__proto__ = b; }) ||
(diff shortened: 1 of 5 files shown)


tests/TestRunner/app/Inheritance/TypeScriptTests.js
```diff
@@ -238,6 +270,78 @@ describe(module.id, function () {
         expect(TNSGetOutput()).toBe('voidSelector called' +
             'variadicSelector:native x:9 called');
     });
+    it("MethodOverrides: errors", function () {
+        expect(function () {
+            var TSObjectErr1 = /** @class */ (function (_super) {
```
vmutafov: You could add the TS code as a comment to see the original code. This is done in the Android Runtime tests and increases the readability.

PR #1230, opened 2019-11-24
fix: don't crash app if flag is present

DickSmith:
## PR Checklist

## What is the current behaviour?
App will crash.

## What is the new behaviour?
App will shutdown.


src/NativeScript/JSErrors.mm
```diff
@@ -181,14 +181,20 @@ void reportFatalErrorBeforeShutdown(ExecState* execState, Exception* exception,
...
-                                                                                               message.utf8().data(),
-                                                                                               jsCallstack.c_str()]
-                                                             reason:nil
-                                                           userInfo:@{ @"sender": @"reportFatalErrorBeforeShutdown" }];
-        @throw objcException;
+        
+        id discardExceptionsValue = [[TNSRuntime current] appPackageJson][@"discardUncaughtJsExceptions"];
+        bool discardExceptions = [discardExceptionsValue boolValue];
+        if (discardExceptions) {
+            exit(0);
```
mbektchiev: I'd like to suggest that you replace the exit code from `0` to some other value to indicate a fatal error. This would help with troubleshooting when debugging issues with such fatal errors. (e.g. some magic value like `139` (coming from `Fatal Error` =  `FE` = 'F'.charCodeAt(0)+'E'.charCodeAt(0) = `139` 😄)

DickSmith: Hey Martin,

So here's the problem. These crashes are luckily relatively low in volume right now (~90 over 2 weeks and and tens of thousands of users), the Crashlytics reports are semi-random so not sure if related,  but the ones sent to Xcode/AppStoreConnect when the user chooses to report them are of this form:

```
Exception Type:  EXC_CRASH (SIGABRT)
Exception Codes: 0x0000000000000000, 0x0000000000000000
Exception Note:  EXC_CORPSE_NOTIFY
Triggered by Thread:  0

Last Exception Backtrace:
0   CoreFoundation                	0x191d6298c __exceptionPreprocess + 220 (NSException.m:199)
1   libobjc.A.dylib               	0x191a8b0a4 objc_exception_throw + 56 (objc-exception.mm:565)
2   NativeScript                  	0x104fde738 NativeScript::reportFatalErrorBeforeShutdown(JSC::ExecState*, JSC::Exception*, bool) + 1432
3   NativeScript                  	0x105022900 -[TNSRuntime executeModule:referredBy:] + 656
4   nativescriptpnp               	0x1041da4d0 0x1041d4000 + 25808
5   libdyld.dylib                 	0x191b65460 start + 4

Thread 0 name:
Thread 0 Crashed:
0   libsystem_kernel.dylib        	0x0000000191b5aebc __pthread_kill + 8
1   libsystem_pthread.dylib       	0x0000000191a76790 pthread_kill$VARIANT$mp + 112 (pthread.c:1456)
2   libsystem_c.dylib             	0x00000001919ca8a0 __abort + 112 (abort.c:147)
3   libsystem_c.dylib             	0x00000001919ca830 abort + 112 (abort.c:118)
4   libc++abi.dylib               	0x0000000191b237d4 abort_message + 128 (abort_message.cpp:76)
5   libc++abi.dylib               	0x0000000191b239c4 demangling_terminate_handler() + 296 (cxa_default_handlers.cpp:66)
6   libobjc.A.dylib               	0x0000000191a8b358 _objc_terminate() + 124 (objc-exception.mm:701)
7   TNSCrashlyticsLogger          	0x0000000106791cb8 CLSTerminateHandler() + 348
8   libc++abi.dylib               	0x0000000191b30304 std::__terminate(void (*)()) + 16 (cxa_handlers.cpp:59)
9   libc++abi.dylib               	0x0000000191b2fc58 __cxxabiv1::failed_throw(__cxxabiv1::__cxa_exception*) + 32 (cxa_exception.cpp:153)
10  libc++abi.dylib               	0x0000000191b2fc18 __cxa_throw + 124 (cxa_exception.cpp:285)
11  libobjc.A.dylib               	0x0000000191a8b1d0 objc_exception_throw + 356 (objc-exception.mm:591)
12  NativeScript                  	0x0000000104fde738 NativeScript::reportFatalErrorBeforeShutdown(JSC::ExecState*, JSC::Exception*, bool) + 1432
13  NativeScript                  	0x0000000105022900 -[TNSRuntime executeModule:referredBy:] + 656
14  nativescriptpnp               	0x00000001041da4d0 0x1041d4000 + 25808
15  libdyld.dylib                 	0x0000000191b65460 start + 4
```
Always the exact same, other than:
1. Address in Thread-0-Frame-4 and Thread-1-Frame-14, obviously
2. Thread-1-Frame-5 is ` demangling_terminate_handler` on iOS 13 and `default_terminate_handler` on iOS 12 (our app supports no older versions, due to low usage)

Other than that, there is no further information and no repro. These are likely extreme edge cases, and account for maybe a ~0.10% crash-rate based on the results from before using this change in our own app vs after. Edge cases can't always be recovered, as may be caused by general instability of the Device/OS (high memory usage, low battery, etc.)

However, just forcing a crash in Production/Release is irresponsible, as it puts pressure on the OS to make sure it is caught, this is usually done by most modern OSes, but is still bad form to just let these things cascade up and rely on the OS to prevent it from going further (and as such, all 1st-party Review and Distributors/Stores are very punitive to this behaviour). As such, since these crashes are so low in volume, and are not reported during development, QA, Customer Service, or even in reviews, there is little value in even reporting them to the OS/1st-party (which an `exit` other than `0` would result in, I believe). All that this results in is bad crash metrics on App Store Connect that I've found results in lower search scores in the App Store.

Technically this flag was already supposed to prevent this, as this method is usually called only in cases where the flag is not enabled, however it seems there are some places it is still being allowed, as such as a final preventative measure I placed it here, under the assumption that maybe there is some cases where the app really should exit. However, I'm not entirely sure if that is necessarily ever the case, and instead debated just suppressing this all together when in Prod.

I did want to confirm that using `exit(0)` results in the crash not being reported, even though the application has been force-exited, though, just for future reference, which is why I tried this solution for now (based on the results of our most recently released app this does seem to be the case, although that was only yesterday, so still monitoring).

The longer term solution may be to allow for 3-4 behaviours here, depending on the value sent to this flag:
1. exception (for debug only);
2. exit with non-0 (better than a crash)
3. `exit(0)`;
4. nothing to keep the app operating (this "fatal" may have occurred during a minor functionality anyway, so it could just better to do nothing and maybe the user not notice than to cause the whole app to terminate which they will). 

**tl;dr** As far as I know, an exit with anything other than 0, while definitely better than throwing an exception up the stack and relying on the OS to stop it, will, as far as I know, be reported to both the OS (and thus the user), and possibly to the App Store if the user chooses to do so, which may have little benefit than hurting the App's score in search results and potential featuring opportunities. I usually have these flags disabled while testing/developing internally, but once in production, app stability is the highest priority.

Also, logging in the cases where this is suppressed is always an option too. As a developer/QA could open the logs for the Released version and even though the app isn't crashing, will be able to see the errors, I believe. With the Crashlytics plugin for {N} all messages from `stderr` are added to any later crash or non-fatal logs (and the uncaught exception handler should be hooked up to Crashlytics for best results).

DickSmith: So just some updates here, after ~24hrs in release and a few thousand users, 99.93% crash free (with only 12 crashes affecting 11 users) on Crashlytics. None yet reported to Xcode (no idea if there is a delay for those ones, since the Crashlytics has been so low already, that's why I started exploring the Xcode/AppStore reports).

Despite being overall lower, there are a couple new ones according to Crashlytics (may be same as some old ones, but with some updates they show up as "different" in the way they group them).

The most curious new one for me is this stack (only 1 so far in Crashlytics):
```
com.google.fira.worker
0  libsystem_kernel.dylib         0x1a827b42c fsync + 8
1  libsqlite3.dylib               0x1aa0fade0 sqlite3_free_table + 63640
2  libsqlite3.dylib               0x1aa0fa8c8 sqlite3_free_table + 62336
3  libsqlite3.dylib               0x1aa0e5d0c sqlite3_value_text + 23504
4  libsqlite3.dylib               0x1aa0b1f44 sqlite3_finalize + 3608
5  libsqlite3.dylib               0x1aa0de774 sqlite3_step + 60828
6  libsqlite3.dylib               0x1aa0cfb5c sqlite3_step + 388
7  libsqlite3.dylib               0x1aa0971cc sqlite3_exec + 300
8  nativescriptpnp                0x104a6fc30 -[APMSqliteStore endTransaction] + 4374428720
9  nativescriptpnp                0x104a6f8e8 -[APMSqliteStore performTransaction:] + 4374427880
10 nativescriptpnp                0x104a48ed0 -[APMMeasurement bundleRawEvents] + 4374269648
11 nativescriptpnp                0x104a41c34 -[APMMeasurement uploadData] + 4374240308
12 nativescriptpnp                0x104a4eb2c __49-[APMMeasurement handleAppWillResignActiveEvent:]_block_invoke_2 + 4374293292
13 nativescriptpnp                0x104a58fc4 __51-[APMScheduler scheduleOnWorkerQueueBlockID:block:]_block_invoke + 4374335428
14 libdispatch.dylib              0x1a814e610 _dispatch_call_block_and_release + 24
15 libdispatch.dylib              0x1a814f184 _dispatch_client_callout + 16
16 libdispatch.dylib              0x1a80fb404 _dispatch_lane_serial_drain$VARIANT$mp + 608
17 libdispatch.dylib              0x1a80fbdf8 _dispatch_lane_invoke$VARIANT$mp + 420
18 libdispatch.dylib              0x1a8105314 _dispatch_workloop_worker_thread + 588
19 libsystem_pthread.dylib        0x1a819ef88 _pthread_wqthread + 276
20 libsystem_pthread.dylib        0x1a81a1ad4 start_wqthread + 8

com.apple.main-thread
0  libsystem_kernel.dylib         0x1a8258634 mach_msg_trap + 8
1  libsystem_kernel.dylib         0x1a8257aa0 mach_msg + 72
2  CoreFoundation                 0x1a83ff04c __CFRunLoopServiceMachPort + 216
3  CoreFoundation                 0x1a83fa16c __CFRunLoopRun + 1444
4  CoreFoundation                 0x1a83f98a0 CFRunLoopRunSpecific + 464
5  GraphicsServices               0x1b2351328 GSEventRunModal + 104
6  UIKitCore                      0x1ac4ea740 UIApplicationMain + 1936
7  NativeScript                   0x1060dc044 ffi_call_SYSV + 68
8  NativeScript                   0x1060d8638 ffi_call_int + 1272
9  NativeScript                   0x1060d8134 ffi_call + 56
10 NativeScript                   0x1055aceb0 NativeScript::FunctionWrapper::call(JSC::ExecState*) + 604
...
...
```
Just as an FYI, the `APM` prefix is for Firebase Analytics, I believe. Now, not 100% sure (as there was about 2-3 weeks of changes between current and last release), but potentially what may have happened here could have been related to the `exit(0)`. It could be that this thread was unable to complete its work while the main was closing down due to something being cleared out too early). 

Still before since main would have already been crashed, this issue likely would have never been the root cause, it's only now occurring perhaps since the `exit(0)` on main has forced it to end too early, or cleared some reference or other thread it was still using.

Like I said, just a guess, and the reason I wanted to try this `exit(0)` stuff to see how the app and Apple crash reporting behave, but I think on the next release I may just suppress this all together, no `exit` and no `throw` which may prevent even more crashes, since these non-main threads may still be able to finish their work gracefully and without crashing, even if the main UI is now unusable for the user (which may itself not really be the case).

I know, officially, Apple prohibits `exit` as part of any standard functionality/behaviour (such as having a quit app button), but they don't seem to completely prohibit any usage, but it's likely still discouraged even for cases like this, where dependant threads may themselves end up behaving badly and cause another crash instead.

So for now I'll continue to monitor, and see if any new crashes are sent to Xcode, but the crash rate is now significantly lower than it was prior to this change (and no other changes in the new release were made to reduce crashes, if anything, I was worried there would be even more new ones). The previous app averaged ~0.15 crash-rate, while this new one is ~0.07 crash-rate on Crashlytics (which means it may or may not have been noticed by the user) and still none in Xcode (which means the user either didn't notice, or didn't choose to report, which is fine either way since it means it was not reported to AppStore).

DickSmith: Per our discussion on the phone, I have made this a combination of #1 and #4. If the `discardUncaughtJsExceptions` is true, the exception is suppressed, else it is thrown.

mbektchiev:
run ci

PR #1275, opened 2020-07-01
fix: invalid isa pointer access in ios 14

rigor789:
## PR Checklist

## What is the new behavior?

Fixes/Implements/Closes #[Issue Number].


rigor789:
@cla-bot check

src/NativeScript/ObjC/IsObjcObject.cpp
```diff
@@ -214,90 +36,13 @@ bool IsObjcObject(const void* inPtr) {
...
-    // Check if the pointer is aligned
-    //
-    if (((uintptr_t)inPtr % sizeof(uintptr_t)) != 0) {
-        return false;
-    }
-
-    //
-    // Check if the memory is valid and readable
-    //
-    if (!IsValidReadableMemory(inPtr)) {
```
mbektchiev, line 236: I think this check needs to be kept because otherwise `object_getClass` will crash in cases where the argument is a value pointing to unreadable memory.

mbektchiev: See:
```
(lldb) p (Class)object_getClass(0x12344)
error: Execution was interrupted, reason: EXC_BAD_ACCESS (code=1, address=0x12344).
The process has been returned to the state before expression evaluation.
```

src/NativeScript/ObjC/IsObjcObject.cpp
```diff
@@ -214,90 +36,13 @@ bool IsObjcObject(const void* inPtr) {
...
-    //
-    if (IsObjcTaggedPointer(inPtr, NULL)) {
-        return true;
-    }
-#endif
-
-    //
-    // Check if the pointer is aligned
-    //
-    if (((uintptr_t)inPtr % sizeof(uintptr_t)) != 0) {
```
mbektchiev, line 229: The check could also be kept as it doesn't use any forbidden APIs and can serve at least as an optimization.

src/NativeScript/ObjC/IsObjcObject.cpp
rigor789: Good point - I wanted to leave it in, but had some trouble with it, specifically this test (and a few more)

```js
    it("Exception in callback propagates to caller and doesn't crash app", function () {
		const jsCallback = TNSPrimitives.extend({
			methodWithNSNumber: function (x) {
				throw new Error("Called with " + x);
			}
		}).alloc().init();

        // performSelector is used to perform a native callback(instead of direct JS call)
		const invocation = () => jsCallback.performSelectorWithObject("methodWithNSNumber:", 127);
        expect(invocation).toThrowError(/Called with 127/);
    });
```

When stepping through with the debugger, `127` would be a NSNumber (or similar, can't recall exactly). `object_getClass` would return the correct class for it, however `IsValidReadableMemory` would return false - thus the whole `IsObjcObject` would return false, and would cause the test to return `null` instead of `127`.

With a bit more testing - seems like previously it was caught by the tagged pointer check.

I'll try to analyze the `IsValidReadableMemory` a bit more, and figure out why `vm_region_64` returns 1 (meaning an error).

src/NativeScript/ObjC/IsObjcObject.cpp
rigor789: This was also giving me trouble with the above case with `127`/`NSNumber`

src/NativeScript/ObjC/IsObjcObject.cpp
mbektchiev: Yes, most probably it was correctly being reported as a tagged pointer. I don't know if this check could also be left there, but if it isn't forbidden by iOS 14 it might be better to keep it as well.

rigor789: Wasn't sure about that either, because the tags have changed in objc too:

From https://opensource.apple.com/source/objc4/objc4-709/runtime/objc-internal.h.auto.html
```c
typedef uint16_t objc_tag_index_t;
enum
#endif
{
    OBJC_TAG_NSAtom            = 0, 
    OBJC_TAG_1                 = 1, 
    OBJC_TAG_NSString          = 2, 
    OBJC_TAG_NSNumber          = 3, 
    OBJC_TAG_NSIndexPath       = 4, 
    OBJC_TAG_NSManagedObjectID = 5, 
    OBJC_TAG_NSDate            = 6, 
    OBJC_TAG_RESERVED_7        = 7, 

    OBJC_TAG_First60BitPayload = 0, 
    OBJC_TAG_Last60BitPayload  = 6, 
    OBJC_TAG_First52BitPayload = 8, 
    OBJC_TAG_Last52BitPayload  = 263, 

    OBJC_TAG_RESERVED_264      = 264
};
```

And the more recent version from https://opensource.apple.com/source/objc4/objc4-781.2/runtime/objc-internal.h.auto.html
```c
typedef uint16_t objc_tag_index_t;
enum
#endif
{
    // 60-bit payloads
    OBJC_TAG_NSAtom            = 0, 
    OBJC_TAG_1                 = 1, 
    OBJC_TAG_NSString          = 2, 
    OBJC_TAG_NSNumber          = 3, 
    OBJC_TAG_NSIndexPath       = 4, 
    OBJC_TAG_NSManagedObjectID = 5, 
    OBJC_TAG_NSDate            = 6,

    // 60-bit reserved
    OBJC_TAG_RESERVED_7        = 7, 

    // 52-bit payloads
    OBJC_TAG_Photos_1          = 8,
    OBJC_TAG_Photos_2          = 9,
    OBJC_TAG_Photos_3          = 10,
    OBJC_TAG_Photos_4          = 11,
    OBJC_TAG_XPC_1             = 12,
    OBJC_TAG_XPC_2             = 13,
    OBJC_TAG_XPC_3             = 14,
    OBJC_TAG_XPC_4             = 15,
    OBJC_TAG_NSColor           = 16,
    OBJC_TAG_UIColor           = 17,
    OBJC_TAG_CGColor           = 18,
    OBJC_TAG_NSIndexSet        = 19,

    OBJC_TAG_First60BitPayload = 0, 
    OBJC_TAG_Last60BitPayload  = 6, 
    OBJC_TAG_First52BitPayload = 8, 
    OBJC_TAG_Last52BitPayload  = 263, 

    OBJC_TAG_RESERVED_264      = 264
};
```

PR #1279, opened 2020-09-29
Update nativescript-build.xcconfig

facetious:
This updates the `EXCLUDED_ARCHS` build variable to remove ARM architectures from the iPhone Simulator build.

I tested this locally with a device attached and a simulator open. Executing `ns run ios` built and ran on both the iPhone and simulator simultaneously.

**Note**: I am not convinced that the additional `ARCH_64` specification needs to be there, but I don't have a 32-bit simulator to build for, so maybe it doesn't even matter anymore?

### DIFF
--- a/build/project-template/internal/nativescript-build.xcconfig
+++ b/build/project-template/internal/nativescript-build.xcconfig
@@ -18,3 +18,6 @@ LDPLUSPLUS = $SRCROOT/internal/nsld.sh
 // * You can learn more about NativeScript metadata in the docs: http://docs.nativescript.org/runtimes/ios/Overview#metadata
 // * It will be generated on each build, so you can find it after running "tns build ios" in "YOUR_APP/platforms/ios".
 // TNS_DEBUG_METADATA_PATH = $(SRCROOT)/debug-metadata
+
+EXCLUDED_ARCHS__PLATFORM_NAME_iphonesimulator__NATIVE_ARCH_64_BIT_x86_64=arm64 arm64e armv7 armv7s armv6 armv8
+EXCLUDED_ARCHS=$(EXCLUDED_ARCHS__PLATFORM_NAME_$(PLATFORM_NAME)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT))


rigor789:
@facetious I was originally thinking to do this in the CLI, but I think adding it here is perhaps easier, and faster to fix if something needs changing down the line. 

If I understand the changes correctly - this will exclude all archs but `x86_64` for the simulator builds - unless the native arch from `NATIVE_ARCH_64_BIT` is not `x86_64` - so in theory on arm macs, it would not exclude anything.

facetious:
@rigor789 Yes, it won't affect the builds for arm macs. My comment was more directed towards the possibility of an i386 existing somewhere that would probably want arm stripped out, too. 🤷

bmacdonald:
@facetious 
This is ignorance on my part.  Can I update the platforms without migrating to nativescript 7?  We are stuck because of the inappbrowser plugin.

rigor789:
@bmacdonald generally yes, but if for some reason - it's not possible, you can add the same lines in your `App_Resources` folder, and build with any runtime - this would just ease the fix without requiring changes to the projects.

bmacdonald:
> @bmacdonald generally yes, but if for some reason - it's not possible, you can add the same lines in your `App_Resources` folder, and build with any runtime - this would just ease the fix without requiring changes to the projects.

@rigor789 
Thank you for the response.  This is how I've been handling the issue with the .xxconfig.  I just never knew if there was a dependency of the platforms on @nativescript/core.

Thank you both for your time.

facetious:
> @rigor789
> Thank you for the response. This is how I've been handling the issue with the .xxconfig. I just never knew if there was a dependency of the platforms on @nativescript/core.
> 
> Thank you both for your time.

The [nativescript-inappbrowser](https://github.com/proyecto26/nativescript-inappbrowser) plugin has been updated to work with `@nativescript/core`, but you'd have to build it locally and link to the file since there's no recent release. You can do this with:

```
%> cd ${your_project_dir}/..
%> git clone git@github.com:proyecto26/nativescript-inappbrowser.git
%> cd nativescript-inappbrowser/src
%> npm i
%> cd ../publish
%> ./pack.sh
%> mkdir ${your_project_dir}/lib
%> mv nativescript-inappbrowser-2.3.0.tar.gz ${your_project_dir}/lib
%> cd ${your_project_dir}
%> npm i lib/nativescript-inappbrowser-2.3.0.tar.gz
```

rigor789 (approved):

rigor789:
I guess we should do the same for the v8 runtime: https://github.com/NativeScript/ns-v8ios-runtime

